PR #24, opened 2022-12-08
Set up test workflow

### LINKED ISSUE #19: Set up automated testing
On a PR against main, the complete test suite should run. [Some inspiration](https://github.com/neurobagel/bagelbids/blob/main/.github/workflows/test.yml)

rmanaem:
The main change is the addition of .github/workflows directory and test.yaml file used to set up test workflow

Closes #19

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -1,0 +1,31 @@
+name: Test
+
+on: [push, pull_request]
+
+jobs:
+  test:
+
+    runs-on: ubuntu-latest
+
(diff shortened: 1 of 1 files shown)


surchs (approved):
Nice, thanks @rmanaem. Looks good to me :tada:

surchs:
Ah one thing to consider: I don't think we need the test-on-push thing? I think I had started that trend without thinking in the bagelbids repo. So maybe remove that event here and then we'll also remove it in the other repos

PR #28, opened 2022-12-13
Implement query parameters for age

### LINKED ISSUE #20: Store test data separately from tests (test_query.py)
Store example data returned from graph in pytest fixture rather than creating it in each test.

### LINKED ISSUE #26: Implement age as a query parameter
- [x] Add `age_min` and `age_max` as optional arguments in `utility.create_query`
- [x] Add " to QueryModel
- [x] Update router and CRUD function for GET path operation to include `age_min` and `age_max`
- [x] Test that valid inputs for min + max age return a 200 status code
- [x] Test that invalid inputs for min + max age returns a 422 status code
  - [x] str input
  - [x] min age > max age
  - [x] min age, max age > 0
  - [ ] ~~min or max age > 100~~

### LINKED ISSUE #29: Rework subject-level filter condition for age in SPARQL query
In `utility.create_query`, if neither the min nor max age are provided, no filter for age is added to the query. However if only one of the values are missing, the missing value is set to a default (0 or 100) and included as a condition in the filter.

A cleaner solution might be to only include filter conditions for non-missing age limits. e.g., if `age=(40, None)` is passed, add a filter only for age >= 40 and not 40 <= age <= 100 by default.

alyssadai:
The main changes are 

- New optional query parameters added: `age_min` and `age_max` with the following constraints:
  - must be floats
  - must be >= 0
  - `age_max` >= `age_min` (via validator added to `QueryModel`)
- Moved example data returned from graph to fixture in `test_query.py`
- Added parameterized tests for valid and invalid age ranges

Closes #20
Closes #26
Closes #29


tests/test_query.py
```diff
@@ -70,48 +76,69 @@ async def mock_get(sex):
...
     response = test_app.get("/query/?sex=apple")
     assert response.status_code == 422
+
+
+@pytest.mark.parametrize(
+    "age_min_keyval, age_max_keyval",
+    [
+        ("age_min=30.5", "age_max=60"),
+        ("age_min=20.75", None),
+        (None, "age_max=50"),
```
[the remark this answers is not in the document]
surchs: Case in point: after reading this test I went to look at the pydantic model and then scratched my head for 30 minutes because I was somehow primed to expect `None` as a valid user supplied value for `age_min` which I now understand is not the case. I do still not understand why you can provide `None` in the query string at all without the whole thing becoming invalid - but that's probably just a limitation of my knowledge of pydantic magic?

surchs (changes requested):
Hey @alyssadai, very cool PR :tada:. I added some change requests for the default values for `age_min` and `age_max`: mainly that they should be set in the pydantic model and not in the query generation method. I think this might also help make existing validation rules a bit clearer to a naive observer (like myself).

tests/test_query.py
```diff
@@ -70,48 +76,69 @@ async def mock_get(sex):
...
+    [
+        ("forty", "fifty"),
+        (33, 21),
+        (-42.5, -40),
+    ],
+)
+def test_get_invalid_age(
+    test_data, test_app, invalid_age_min, invalid_age_max, monkeypatch
+):
+    """Given a valid min age and max age, returns a 200 status code and a non-empty list of results."""
```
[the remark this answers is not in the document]
alyssadai: Oops, yes! Good catch

rmanaem (changes requested):
Great work as always Alyssa :fire:
I added some comments but the change request is regarding the `test_get_invalid_age`.

app/api/utility.py
```diff
@@ -33,28 +33,47 @@
...
 
     Returns
     -------
     str
         The SPARQL query.
     """
     subject_level_filters = ""
-    session_level_filters = ""
+
+    if (
```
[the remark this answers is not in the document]
rmanaem: Yea I agree, I think that snippet is unnecessary.

app/api/models.py
```diff
@@ -11,16 +11,24 @@ class QueryModel(BaseModel):
     """Data model and dependency for API that stores the query parameters to be accepted and validated."""
 
     sex: Literal["male", "female", "other"] = None
-    age_min: float = Query(default=None, gt=0)
-    age_max: float = Query(default=None, gt=0)
+    age_min: float = Query(default=None, ge=0)
+    age_max: float = Query(default=None, ge=0)
```
surchs, line 15: good catch!

app/api/models.py
```diff
@@ -11,16 +11,24 @@ class QueryModel(BaseModel):
...
+    def check_agemax_ge_agemin(cls, values):
+        """
+        If both age bounds have been set to values other than their defaults (None), ensure that age_max is >= age_min.
+        NOTE: HTTPException (and not ValueError) raised here to get around "Internal Server Error" raised by
+        FastAPI when a validation error comes from a Pydantic validator inside a class dependency.
+        See:
+        https://github.com/tiangolo/fastapi/issues/1474
+        https://github.com/tiangolo/fastapi/discussions/3426
+        https://fastapi.tiangolo.com/tutorial/handling-errors/?h=validation#requestvalidationerror-vs-validationerror
+        """
```
surchs, line 27: :+1: very clear!

tests/test_query.py
```diff
@@ -100,23 +100,39 @@ async def mock_get(age_min, age_max, sex):
...
 ):
-    """Given a valid min age and max age, returns a 200 status code and a non-empty list of results."""
+    """Given only a valid lower/upper age bound, returns a 200 status code and a non-empty list of results."""
 
     async def mock_get(age_min, age_max, sex):
         return test_data
 
     monkeypatch.setattr(crud, "get", mock_get)
-    response = test_app.get(f"/query/?{age_min_keyval}&{age_max_keyval}")
+    response = test_app.get(f"/query/?{age_keyval}")
```
surchs, line 135: aha, that's a great solution! :tada:

app/api/utility.py
```diff
@@ -51,22 +51,11 @@ def create_query(age: tuple = (None, None), sex: str = None) -> str:
...
-            (
-                default_val if age_val is None else age_val
-                for age_val, default_val in zip(age, [0, 100])
-            )
-        )
-        subject_level_filters += (
-            "\n"
-            + f"FILTER (?{AGE.var} >= {age[0]} && ?{AGE.var} <= {age[1]})."
-        )
+    if isinstance(age, tuple):
```
surchs, line 54: what should happen if `age` is not a tuple? Probably out of scope for this PR. But with the current setup, we could have a silent failure if we somehow pass a non-tuple (e.g. another iterable like a list) by accident. Then we think we filter age, but we don't. One option would be to remove the `if` here and accept that we may get an `IndexError` if we somehow just pass an `int`. Another option would be to manually raise a `TypeError` if it is not `tuple`. 

Afaik, only our own code will ever call this function. So we could decide to not handle "user error" gracefully and instead just crash. Then we'll probably catch a mistake like "we thought we could pass `age=12` to `create_query` and get away with it" because our tests will be failing (hopefully). 

We should probably do this in a separate issue though because the code here is working and we don't want to drag this out!

surchs (approved):
Great work @alyssadai. I left a comment for a later iteration but this is good to go in now! :tada:

PR #39, opened 2022-12-16
Implemented image modality query parameter

### LINKED ISSUE #23: Implement image modality query parameter
- [x] Implement image modality in create_query function of utility.py module
- [x] Add image modality as a field in QueryModel of models.py module
- [x] Update get_query path operation of query.py router and get function of crud.py module to reflect the addition of image modality
- [x] test parameter implementation/functionality

rmanaem:
Closes #23


app/api/utility.py
```diff
@@ -32,17 +32,31 @@
 
 CATEGORICAL_DOMAINS = [SEX, DIAGNOSIS, IMAGE_MODAL]
 
+# Query choices
+IMAGE_MODAL_CHOICES = {
+    "All": None,
```
alyssadai, line 37: Do we necessarily need to include `image_modal="All"` as an option if the behaviour is (from what I can tell) the same as when you don't filter for age in your query at all, since this is an optional parameter?

tests/test_query.py
```diff
@@ -150,11 +150,53 @@ def test_get_invalid_age(
...
+
+    monkeypatch.setattr(crud, "get", mock_get)
+    response = test_app.get(f"/query/?image_modal={valid_image_modal}")
+    assert response.status_code == 200
+    assert response.json() != []
+
+
+@pytest.mark.parametrize(
+    "invalid_image_modal",
+    ["diffusion weighted", "eeg", "Flow weghed", "T1weighted", "apple"],
```
alyssadai, line 190: This is a bit picky and I'll defer to @surchs' opinion here, but I would lean towards not having lowercase image modality names as 'invalid' test cases. I think in the case of an isolated API, we would want str params to be case-insensitive (i.e., both "diffusion weighted" and "Diffusion weighted" should be valid inputs). Obviously, since we'll soon have a GUI frontend that presumably has exact options the user can select rather than accepting any str, I don't think we urgently need to implement case-insensitivity right now - but maybe it would still be better conceptually to avoid having them in the codebase as explicit examples of invalid parameters?

alyssadai (commented):
Nicely done @rmanaem! I found your PR very concise and easy to follow. 🎉
On my end I have just a couple minor conceptual comments below.

tests/test_query.py
rmanaem: I was hoping this would come up so thanks for mentioning it. I wasn't sure how restrictive we'd wanna be with the input so I decided to go with the most restrictive and intentionally mentioned those cases in test for demonstration.
I'd be happy to implement case insensitivity if we end up going that way.

app/api/utility.py
rmanaem: I agree. I only kept it in case we need it later in the frontend.

surchs (commented):
Nice job @rmanaem :tada:! Code looks great. My main comment is: do we really need to do all this user-handholding with query parameter validation and matching string labels to terms? I know we discussed this - and apologies for bringing this up after you have done the implementation. But the goal for us is to get this thing deployed fast, and to do so with as little complexity as we can get away with while still being useful. Especially since the first "user" will be another of our own tools, I think just expecting and passing along the actual term IRIs is more maintainable for categorical variables (with the possible exception of sex). 

what do you and @alyssadai think?

app/api/models.py
```diff
@@ -6,13 +6,16 @@
...
+
 
 class QueryModel(BaseModel):
     """Data model and dependency for API that stores the query parameters to be accepted and validated."""
 
-    sex: Literal["male", "female", "other"] = None
     age_min: float = Query(default=None, ge=0)
     age_max: float = Query(default=None, ge=0)
+    sex: Literal["male", "female", "other"] = None
+    image_modal: Literal[tuple(IMAGE_MODAL_CHOICES.keys())] = None
```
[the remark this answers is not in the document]
rmanaem: The input must be immutable/hashable

PR #41, opened 2022-12-21
Implement query parameters for diagnosis and is_control

### LINKED ISSUE #30: Remove check for tuple age in create_query function
There is currently an if in create_query that checks whether age is a tuple before proceeding to add age filter conditions.
https://github.com/neurobagel/bagelapi/blob/cd15c2eef1b51a683f60d8d221b5009376d1bc9b/app/api/utility.py#L54-L58

To avoid silent failures if a non-tuple age argument is passed by accident (e.g., by our own code), this if could be removed to expose IndexErrors when this happens.

### LINKED ISSUE #32: Implement diagnosis and is_control as query parameters
- [x] Implement diagnosis (str) and is_control (bool) in create_query function of utility.py module
- [x] Add diagnosis and is_control as a field in QueryModel of models.py module
- [x] Update get_query path operation of query.py router and get function of crud.py module to reflect the addition of diagnosis and is_control
- [x] Test that the following return 200 status codes:
  - [x] valid diagnosis term
  - [x] `is_control = True` and `is_control = False`
- [ ] ~~Test that unavailable diagnosis term returns empty list + 200 status code~~
- [x] Test that the following return error codes:
   - [x] invalid diagnosis term, 422
   - [x] valid diagnosis + `is_control = True`, 422

### LINKED ISSUE #40: Create pytest fixture for mock get function
A `mock_get` function with the same args and returning the same test data is currently re-defined in each test for a valid query parameter. For code conciseness, this function could be defined once in a fixture and monkeypatched inside each test, as described here:

https://stackoverflow.com/questions/50842956/how-to-monkey-patch-a-function-for-multiple-tests

Optional: A separate fixture could also be created for a mock get function that returns None (for testing invalid inputs).

alyssadai:
The main changes are:

- New optional query parameters added: diagnosis (str) and is_control (bool) 
- New validator added in QueryModel to ensure diagnosis and is_control values are mutually exclusive
- Created pytest fixture for mock_successful_get function
- Added tests for valid and invalid diagnosis and/or is_control inputs

Closes #32
Closes #40
Closes #30


app/api/models.py
```diff
@@ -33,3 +35,12 @@ def check_agemax_ge_agemin(cls, values):
                 detail="'age_max' must be greater than or equal to 'age_min'",
             )
         return values
+
+    @root_validator
+    def check_exclusive_diagnosis_or_ctrl(cls, values):
+        if values["diagnosis"] is not None and values["is_control"]:
```
surchs, line 41: nice!

app/api/utility.py
```diff
@@ -55,16 +65,29 @@ def create_query(
     """
     subject_level_filters = ""
 
-    if isinstance(age, tuple):
-        if age[0] is not None:
-            subject_level_filters += "\n" + f"FILTER (?{AGE.var} >= {age[0]})."
-        if age[1] is not None:
-            subject_level_filters += "\n" + f"FILTER (?{AGE.var} <= {age[1]})."
+    if age[0] is not None:
```
surchs, line 68: this is good, I agree with doing it. But I think in the future it'll be faster to do this in a separate PR because it is unrelated to the diagnosis / control_subject context here and those are already a bit complex.

app/api/utility.py
```diff
@@ -55,16 +65,29 @@ def create_query(
...
+            "\n" + f"FILTER (?{DIAGNOSIS.var} = {diagnosis})."
+        )
+
+    if is_control is not None:
+        if is_control:
+            subject_level_filters += (
+                "\n" + f"FILTER (?{IS_CONTROL.var} = <{IS_CONTROL_TERM}>)."
+            )
+        else:
+            subject_level_filters += (
```
surchs, line 87: oh, interesting. I hadn't even thought of that. This is the "give me only patients" use case? Good idea.

This also looks good to me, I'll just leave a note that [SPARQL has a negation pattern](https://www.w3.org/TR/sparql11-query/#negation) for the `FILTER` command and I am not completely sure if the two are equivalent when there is more than one relation of the same type (e.g. if you somehow end up with several `bg:isSubjectGroup` statements, one of which meets your Filter condition here.

We should make sure we understand what the differences are between the two.

tests/test_query.py
```diff
@@ -232,11 +235,83 @@ def test_get_undefined_prefix_image_modal(
...
+    async def mock_get(
+        age_min, age_max, sex, diagnosis, is_control, image_modal
+    ):
+        return None
+
+    monkeypatch.setattr(crud, "get", mock_get)
+    response = test_app.get(
+        "/query/?diagnosis=snomed:35489007&is_control=True"
+    )
+    assert response.status_code == 422
```
surchs, line 317: All of these look great. I would also assert that the response has the new message about subject and is_control not both being allowed to be set.

surchs (approved):
Looks great @alyssadai! I left some small comments - otherwise good to go!

app/api/utility.py
alyssadai: Noted, will do!

tests/test_query.py
```diff
@@ -210,13 +213,13 @@ async def mock_get(age_min, age_max, sex, image_modal):
...
-):
+def test_get_invalid_image_modal(test_app, invalid_image_modal, monkeypatch):
     """Given an invalid image modality, returns a 422 status code."""
 
-    async def mock_get(age_min, age_max, sex, image_modal):
-        return test_data
+    async def mock_get(
+        age_min, age_max, sex, diagnosis, is_control, image_modal
+    ):
+        return None
```
rmanaem, line 222: Thanks for fixing this, it seems I forgot.

rmanaem (approved):
Looks good to me!

app/api/utility.py
alyssadai: Interesting, thanks for sharing! I've made a note of this.

PR #46, opened 2022-12-27
Implemented minimum number of imaging sessions query parameter

### LINKED ISSUE #34: Implement a minimum number of (imaging) sessions as a query parameter
- [x] Implement minimum number of sessions in create_query function of utility.py module
- [x] Add minimum number of sessions as a field in QueryModel of models.py module
- [x] Update get_query path operation of query.py router and get function of crud.py module to reflect the addition of minimum number of sessions
- [x] test parameter implementation/functionality

rmanaem:
Closes #34

### DIFF
--- a/app/api/utility.py
+++ b/app/api/utility.py
@@ -22,7 +22,7 @@ PREFIX nidm: <http://purl.org/nidash/nidm#>
 # Store domains in named tuples
 Domain = namedtuple("Domain", ["var", "pred"])
 # Core domains
-AGE = Domain("age", "bg:hasAge")
+AGE = Domain("age", "bg:age")
 SEX = Domain("sex", "bg:sex")
 DIAGNOSIS = Domain("diagnosis", "bg:diagnosis")
 IS_CONTROL = Domain("subject_group", "bg:isSubjectGroup")
--- a/app/api/utility.py
+++ b/app/api/utility.py
@@ -25,10 +25,11 @@ Domain = namedtuple("Domain", ["var", "pred"])
 AGE = Domain("age", "bg:hasAge")
 SEX = Domain("sex", "bg:sex")
 DIAGNOSIS = Domain("diagnosis", "bg:diagnosis")
+IS_CONTROL = Domain("subject_group", "bg:isSubjectGroup")
 IMAGE_MODAL = Domain("image_modal", "bg:hasContrastType")
 TOOL = Domain("tool", "")
 PROJECT = Domain("project", "bg:hasSamples")
-IS_CONTROL = Domain("subject_group", "bg:isSubjectGroup")
+
 
 CATEGORICAL_DOMAINS = [SEX, DIAGNOSIS, IMAGE_MODAL]
 
@@ -40,6 +41,7 @@ def create_query(
     sex: str = None,
     diagnosis: str = None,
     is_control: bool = None,
+    min_num_sessions: int = None,
     image_modal: str = None,
 ) -> str:
     """
@@ -55,6 +57,8 @@ def create_query(
         Subject diagnosis, by default None.
     is_control : bool, optional
         Whether or not subject is a control, by default None.
+    min_num_sessions : int, optional
+        Subject minimum number of imaging sessions, by default None.
     image_modal : str, optional
         Imaging modality of subject scans, by default None.
 
@@ -88,6 +92,11 @@ def create_query(
                 "\n" + f"FILTER (?{IS_CONTROL.var} != <{IS_CONTROL_TERM}>)."
             )
 
+    if min_num_sessions is not None:
+        subject_level_filters += (
+            "\n" + f"FILTER (?num_sessions >= {min_num_sessions})."
+        )
+
     session_level_filters = ""
 
     if image_modal is not None:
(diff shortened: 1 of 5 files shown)


app/api/utility.py
```diff
@@ -40,6 +41,7 @@ def create_query(
     sex: str = None,
     diagnosis: str = None,
     is_control: bool = None,
+    min_num_sessions: int = None,
```
surchs, line 44: I think the correct way here would be to use 

```python
from typing import Optional
min_num_sessions: Optional[int] = None
```

because we are using type hinting and `None` as a default for an `int`. See here: https://docs.python.org/3/library/typing.html#typing.Optional

This is true of the other stuff as well, and in other places, so maybe that's something for another PR.

tests/test_query.py
```diff
@@ -42,7 +42,13 @@ def mock_successful_get(test_data):
     """Mock get function that returns non-empty query results."""
 
     async def mock_ret(
```
surchs, line 44: Was this supposed to be called `..._get`?

tests/test_query.py
```diff
@@ -319,3 +273,161 @@ async def mock_get(
...
+
+    monkeypatch.setattr(crud, "get", mock_successful_get)
+    response = test_app.get(
+        f"/query/?min_num_sessions={valid_min_num_sessions}"
+    )
+    assert response.status_code == 200
+    assert response.json() != []
+
+
+def test_get_valid_nonsensical_min_num_sessions(test_app, monkeypatch):
```
surchs, line 292: I don't understand this test. Why should that fail?
What if Russ Poldrack wants to put all of his myconnectome data into Neurobagel and model them as sessions: http://www.myconnectome.org/ ? I think any valid number (e.g. positive integer) is also sensible.

tests/test_query.py
```diff
@@ -319,3 +273,161 @@ async def mock_get(
...
+    response.json() == []
+
+
+@pytest.mark.parametrize("invalid_min_num_sessions", [0, -3, "apple"])
+def test_get_invalid_min_num_sessions(
+    test_app, invalid_min_num_sessions, monkeypatch
+):
+    """Given an invalid minimum number of imaging sessions, returns a 422 status code."""
+
+    async def mock_get(
```
surchs, line 318: sorry for bringing this all up here, but I don't think we need to even mock the get method for invalid parameters. If the internal validation does its job, then there shouldn't be any get request going out that we would have to mock. So we could just call the API without the mock and expect it to fail. That would also make the test more legible, becuase we're not asserting over the returned result list being empty (because that doesn't make sense here), so it's not immediately clear why we are mocking the getter to return an empty list.

surchs: Because you have already reshuffled some of the tests here, I think it would be fine for you to also get rid of the "unhappy-path-test-get-mock" stuff.

surchs (changes requested):
Nice PR @rmanaem, thanks! I left some minor comments. My main thing is: why should we differentiate between sensible and non-sensible numbers of sessions as long as they are valid.

I also noticed a couple of things that aren't a new change from your PR but kinda fit in because of other changes you made. We can also address these in a separate PR.

tests/test_query.py
rmanaem: I think so but I'll check with Alyssa to be sure.

app/api/utility.py
rmanaem: Yea I think it's better to correct this in another issue.

tests/test_query.py
rmanaem: No thanks for bringing it up, I've had this question in my head for a while but somehow forgot to look into it or even ask about it.
I can go through the tests and do a refactor.

tests/test_query.py
rmanaem: I see your point

alyssadai (commented):
Thanks for the PR @rmanaem! I've added to a few of @surchs' comments but otherwise have no major suggestions. I noticed a tiny typo (currently harmless) that could also be fixed in this PR, but could also be addressed separately.

app/api/utility.py
```diff
@@ -25,10 +25,11 @@
 AGE = Domain("age", "bg:hasAge")
```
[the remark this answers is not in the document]
rmanaem: I think it's fine to fix in this PR.

tests/test_query.py
rmanaem: I think renaming it to mock_return will clear things up

tests/test_query.py
```diff
@@ -319,3 +273,141 @@ async def mock_get(
         "Subjects cannot both be healthy controls and have a diagnosis"
         in response.text
     )
+
+
+@pytest.mark.parametrize("valid_min_num_sessions", [1, 2, 4, 7])
+def test_get_valid_sensible_min_num_sessions(
```
surchs, line 279: @rmanaem: can you rename this thing (and docstring) and remove "sensible" here as well, for consistency?

surchs (approved):
@rmanaem good to go, thanks! Before you merge: can you please remove the "sensible" part of the passing test name and description as well - to be consistent.

rmanaem:
@surchs Sure, I think this is a good case for a force push. What do you think?

surchs:
> @surchs Sure, I think this is a good case for a force push. What do you think?

Why do you think so? Can't you just add this as a new commit to the branch? I think this whole PR will be squash merged at the end anyway, but I am not sure I understand why you'd need to force push here.

surchs:
Ah ok, didn't realize that you were merging. Yeah, sure, if you'd like to rebase before merge that's all good by me!

PR #47, opened 2023-01-03
Updated README.md and Modified Workflow Trigger Events

### LINKED ISSUE #21: README.md
The README should have information on how one can:

- [x] setup a local dev environment (venv + pip)
- [x] launch the API (uvicorn ... + set environment variables)
- [x] how to run the pytest tests (env variables are set + the server is reachable?)

### LINKED ISSUE #42: Only run test suite in pull requests
Our excellent test suite currently [runs on two events](https://github.com/neurobagel/bagelapi/blob/main/.github/workflows/test.yaml): push and pull_request ([see also](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows)). I think they should only run for pull _request so we don't run them twice during a PR.

rmanaem:
Closes #21 
Closes #42

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,3 @@
-[//]: <> (Do we need to include a section explaining the server connection whether that'd be McGill vpn or something else?)
 <div align="center">
 
 # Bagel API
@@ -12,7 +11,7 @@
         <img src="https://img.shields.io/coverallsCoverage/github/neurobagel/bagelapi?style=flat-square&color=8FBC8F">
     </a>
     <a href="https://www.python.org/">
-        <img src="https://img.shields.io/badge/python-3.7%20%7C%203.8%20%7C%203.9%20%7C%203.10-4682B4?style=flat-square" alt="Python">
+        <img src="https://img.shields.io/badge/python-3.10-4682B4?style=flat-square" alt="Python">
     </a>
     <a href="LICENSE">
         <img src="https://img.shields.io/github/license/neurobagel/bagelapi?color=CD5C5C&style=flat-square" alt="GitHub license">
--- a/README.md
+++ b/README.md
@@ -1,1 +1,98 @@
+[//]: <> (Do we need to include a section explaining the server connection whether that'd be McGill vpn or something else?)
+<div align="center">
+
 # Bagel API
+
+    
+<div>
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -1 +1,98 @@
+[//]: <> (Do we need to include a section explaining the server connection whether that'd be McGill vpn or something else?)
```
surchs, line 1: is this leftover from our discussion?

app/api/models.py
```diff
@@ -1,10 +1,9 @@
 """Data models."""
 
-from typing import Literal
-
 from fastapi import Query
 from fastapi.exceptions import HTTPException
 from pydantic import BaseModel, constr, root_validator
+from typing_extensions import Literal
```
surchs, line 6: this may be a good idea (I don't know), but let's handle it in a different PR so we don't confuse ourselves later :)

surchs (changes requested):
Thanks a lot for this PR, @rmanaem :tada:! Very nice README file! I flagged two things that seem to have accidentally snuck into the PR, but otherwise good to go!

README.md
rmanaem: Yea I left this as a reminder for the discussion.

app/api/models.py
rmanaem: It is in a way related to the README.md file update. While I was adding badges, I updated the test suite to run for Python 3.7-3.10. The suite for 3.7 failed due to a discrepancy in the Literal package import statement.

surchs: ah, great thinking! That's good to check. But also a big enough thing that we should make a separate issue to discuss what versions we want to support and then be intentional about that. So I'd remove it here and bring it back when we address that.

README.md
surchs: OK. But we wrapped up the discussion and said we wouldn't include it for now, no? So I'd remove it here so we don't wonder the same thing every time we see the raw text.

rmanaem: I think we decided to have a discussion in PR(or draft) instead but anyways I had this there before bringing it up in the issue. I'll take care of it.

surchs (approved):
looks great! Thanks @rmanaem :tada: gtg

Issue #48, opened 2023-01-04
Refactor get method implementation


surchs:
I think we could probably rethink the way the `get` method works overall. I think right now it does too many things:

1. Act as a wrapper to the `create_query` method by passing along all arguments intended for that guy
2. Act as a wrapper for the `httpx.post` call by handling the header stuff
3. as an error handler for the post call

We should:
- ~get rid of `get`, put all the logic inside the query router~
- ~call the query generator outside of `get`/`post`, store the output and pass that along~
- [x] make a slimmed down version of a `post` wrapper that receives the completed query rather than the whole argument chain
- [x] refactor creation of the SPARQL query context into its own function

alyssadai:
This is a fair point - thanks @surchs! A couple initial thoughts from me:

I think the fact that the `get` method does 2. and 3. is not *too* far-fetched - I've seen in tutorials examples where something similar is done, but with the [`get` method as a wrapper for a call to a SQL database](https://testdriven.io/blog/fastapi-crud/#code_1). I do see that the lack of tests for the `create_query` function is not ideal, as I suppose we don't check explicitly that the generated query is well-formed aside from Stardog not complaining at us when it's sent through - since we don't have the help of a SPARQL engine to format queries.

In terms of your ideas for simplifying `get`, I'm leaning more towards either the second or third option since I believe having the CRUD operations separated from the path operations (i.e., `@router.get(...)`/`@app.get(...)`) is generally recommended, especially for the purpose of unit tests in more complex apps (see https://fastapi.tiangolo.com/tutorial/sql-databases/#crud-utils and the Tip under the "Read data" code snippet).

I do think we could also (?) have a unit test for the `create_query` function - I could see this being stored in a separate test module (test_commands.py?), along with any future tests for Stardog post requests (e.g., a second template containing INSERT statements for updating the graph).

PR #55, opened 2023-01-11
Modify query path to return only aggregate subject info

### LINKED ISSUE #49: Return the name (and identifier) of each dataset that has matching subjects for a given query

### LINKED ISSUE #50: Return the number of subjects that match a query
- [x] Modify utility.create_query to return only dataset-level info and to group results by unique `sub_id`
- [x] Create a response model for aggregate response data
- [x] Modify crud.get to return total number of matching subjects rather than subject-level dicts

### LINKED ISSUE #52: Update tests to use aggregate response model
- [x] Update test data to reflect aggregate response model
- [x] Update tests expecting a successful query but no query results

alyssadai:
The main changes are:

- Modified SPARQL query to not return any subject-level phenotypic info
- Created + implemented data model for aggregate responses (dataset, dataset name, number of matching subjects)
- Updated tests to use aggregate responses

Closes #50
Closes #49 
Closes #52

### DIFF
--- a/app/api/crud.py
+++ b/app/api/crud.py
@@ -75,23 +75,27 @@ async def get(
     ]
     results_df = pd.DataFrame(results_dicts)
 
-    agg_dicts = (
-        []
-        if results_df.empty
-        else results_df.groupby(by=["dataset", "dataset_name"])
-        .size()
-        .reset_index(name="subjects")
-        .sort_values(by=["dataset_name"])
-        .to_dict("records")
-    )
-
-    response_obj = [
-        AggDatasetResponse(
-            dataset=i["dataset"],
-            dataset_name=i["dataset_name"],
-            num_matching_subjects=i["subjects"],
+    if results_df.empty:
+        response_obj = [
+            AggDatasetResponse(
+                num_matching_subjects=results_df.squeeze().shape[0]
+            )
+        ]
+    else:
+        agg_dicts = (
+            results_df.groupby(by=["dataset", "dataset_name"])
+            .size()
+            .reset_index(name="num_subjects")
+            .sort_values(by="dataset_name")
+            .to_dict("records")
         )
-        for i in agg_dicts
-    ]
+        response_obj = [
+            AggDatasetResponse(
+                dataset=i["dataset"],
+                dataset_name=i["dataset_name"],
+                num_matching_subjects=i["num_subjects"],
+            )
+            for i in agg_dicts
+        ]
 
     return response_obj
--- a/app/api/crud.py
+++ b/app/api/crud.py
@@ -68,34 +68,23 @@ async def get(
 
     results = response.json()
 
-    # return results
     results_dicts = [
         {k: v["value"] for k, v in res.items()}
         for res in results["results"]["bindings"]
     ]
     results_df = pd.DataFrame(results_dicts)
 
-    if results_df.empty:
-        response_obj = [
-            AggDatasetResponse(
-                num_matching_subjects=results_df.squeeze().shape[0]
+    response_obj = []
+    if not results_df.empty:
+        for (dataset, dataset_name), group in results_df.groupby(
+            by=["dataset", "dataset_name"]
+        ):
+            response_obj.append(
+                AggDatasetResponse(
+                    dataset=dataset,
+                    dataset_name=dataset_name,
+                    num_matching_subjects=group.shape[0],
+                )
             )
-        ]
-    else:
-        agg_dicts = (
-            results_df.groupby(by=["dataset", "dataset_name"])
-            .size()
-            .reset_index(name="num_subjects")
-            .sort_values(by="dataset_name")
-            .to_dict("records")
-        )
-        response_obj = [
-            AggDatasetResponse(
-                dataset=i["dataset"],
-                dataset_name=i["dataset_name"],
-                num_matching_subjects=i["num_subjects"],
-            )
-            for i in agg_dicts
-        ]
 
     return response_obj
--- a/app/api/crud.py
+++ b/app/api/crud.py
@@ -75,9 +75,23 @@ async def get(
     ]
     results_df = pd.DataFrame(results_dicts)
 
-    # .squeeze() is needed below for case where results_dict contains a [{}] (GROUP BY behaviour)
+    agg_dicts = (
+        []
+        if results_df.empty
+        else results_df.groupby(by=["dataset", "dataset_name"])
+        .size()
+        .reset_index(name="subjects")
+        .sort_values(by=["dataset_name"])
+        .to_dict("records")
+    )
+
     response_obj = [
-        AggDatasetResponse(num_matching_subjects=results_df.squeeze().shape[0])
+        AggDatasetResponse(
+            dataset=i["dataset"],
+            dataset_name=i["dataset_name"],
+            num_matching_subjects=i["subjects"],
+        )
+        for i in agg_dicts
     ]
 
     return response_obj
--- a/app/api/crud.py
+++ b/app/api/crud.py
@@ -2,9 +2,11 @@
 import os
 
 import httpx
+import pandas as pd
 from fastapi import HTTPException
 
 from . import utility as util
+from .models import AggDatasetResponse
 
 
 async def get(
@@ -66,7 +68,16 @@ async def get(
 
     results = response.json()
 
-    return [
+    # return results
+    results_dicts = [
         {k: v["value"] for k, v in res.items()}
         for res in results["results"]["bindings"]
     ]
+    results_df = pd.DataFrame(results_dicts)
+
+    # .squeeze() is needed below for case where results_dict contains a [{}] (GROUP BY behaviour)
+    response_obj = [
+        AggDatasetResponse(num_matching_subjects=results_df.squeeze().shape[0])
+    ]
+
+    return response_obj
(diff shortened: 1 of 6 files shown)


app/api/models.py
```diff
@@ -45,3 +45,11 @@ def check_exclusive_diagnosis_or_ctrl(cls, values):
...
             )
         return values
+
+
+class AggDatasetResponse(BaseModel):
+    """Data model for query results aggregated at the dataset-level."""
+
+    dataset: str = None
+    dataset_name: str = None
+    num_matching_subjects: int
```
surchs, line 55: Is the number of subjects the only mandatory attribute here?

tests/test_query.py
```diff
@@ -83,12 +70,12 @@ def mock_httpx_post(**kwargs):
...
 
 def test_get_all(test_app, mock_successful_get, monkeypatch):
-    """Given no input for the sex parameter, returns a 200 status code and a non-empty list of results (should correspond to all subjects in graph)."""
+    """Given no input for the sex parameter, returns a 200 status code and at least one dataset with subjects (should correspond to all subjects in graph)."""
 
     monkeypatch.setattr(crud, "get", mock_successful_get)
     response = test_app.get("/query/")
     assert response.status_code == 200
-    assert response.json() != []
+    assert 0 not in [i["num_matching_subjects"] for i in response.json()]
```
surchs, line 78: I like the more explicit test. But I find this assertion confusing. The `"num_matching_subjects"` thing is coming from `get`, but we know that this is mocked out. So you are essentially just asserting that the mocked data fixture from above has a key for `num_matching...` and that its value is not 0.

In this case I think it makes more sense to keep the previous assertion because that just checks that something comes back out of the app. We can also talk through the tests a bit more in https://github.com/neurobagel/bagelapi/issues/48 or something related.

surchs (changes requested):
Very cool PR @rmanaem and @alyssadai :tada:. 

I left two comments. One for a section of code that I think could be a bit clearer, and another about the tests. The test one is more of a question. Please address the clarity one for `crud` and then this is good to go!

app/api/crud.py
```diff
@@ -66,7 +68,34 @@ async def get(
 
     results = response.json()
 
-    return [
+    # return results
```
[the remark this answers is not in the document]
alyssadai: I think we can remove this comment

app/api/crud.py
```diff
@@ -66,7 +68,34 @@ async def get(
...
+    results_dicts = [
         {k: v["value"] for k, v in res.items()}
         for res in results["results"]["bindings"]
     ]
+    results_df = pd.DataFrame(results_dicts)
+
+    if results_df.empty:
+        response_obj = [
+            AggDatasetResponse(
+                num_matching_subjects=results_df.squeeze().shape[0]
```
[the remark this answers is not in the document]
alyssadai: > Wouldn't results_df only be empty if there are zero matching subjects?

Yes, I think you're right. I considered just having `num_matching_subjects=0` here, but I figured _if_ for whatever reason an empty list was returned where it wasn't expected, that would be caught by computing the number this way. This may be overcomplicating things though, so we can just hard code it.

> Related: could we just return an empty list in this case? Why does there have to be one (empty) return object?

This is an option. @rmanaem & I were honestly a bit conflicted about what to return when 0 subjects match the query, but we ended up deciding to be consistent with the aggregate response data model since `num_matching_subjects` is available - why wouldn't we take advantage of it in this case? With the current implementation, the response to the user for a well-formed query that doesn't have matching subjects is:
```
{
[
  "dataset": null,
  "dataset_name": null,
  "num_matching_subjects": = 0,
]
}
```
which I think is clearer and more consistent with the format of successful queries than if just `{[]}` was returned.
What do you think @surchs?

app/api/models.py
alyssadai: For now, yes, since we are accounting for the case where we want to return to the user that `num_matching_subjects = 0`, in which case `dataset` and `dataset_name` wouldn't have a value.

tests/test_query.py
rmanaem: I don't think having the old version makes sense in the current implementation, the response is going to be a list either way. We wanted to differentiate between a (happy) test with no matching data and one with matching data e.g.

https://github.com/neurobagel/bagelapi/blob/aab9571e865f1822b50451e915203c2033bbb009/tests/test_query.py#L353 

Here we expect no matching subjects (the old empty list) as a response so we check if the number of matching subjects is zero for the only dictionary in the returned list.
Another reason I had in mind was to in a way document our intention with the current implementation and our expectation of how the API would respond i.e. a list containing dictionaries (corresponding to each dataset) none of which have 0 `num_matching_subjects`.

app/api/crud.py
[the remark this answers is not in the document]
surchs: Ah I see. So right now the API will always return a list of aggDataset objects, correct? That is:

- if I find something, I get a list of (agg)Dataset objects for those datasets in the graph that have any matches
- if I change my criteria, the number of returned dataset objects might change (and also the number of matched subjects in each will change)
- If I find nothing, I will get only one dataset back, but it's not really a dataset, more of a ghost-dataset with no name and zero matching subjects. When I get that, I know that nothing has matched.

I think that last rule is a bit confusing. If you give me an empty ghost-dataset if there are no matches in ANY datasets, why don't you just do this for ALL datasets, i.e. always return a list of all datasets in the graph and only set the num_matches to zero for the ones without matches. To be clear, I don't think that'd better, just more consistent.

To me an empty list would be clear: nothing found. But if you don't like that, you could also make a global response object and nest the dataset objects inside of it. So 

```python
class Response(BaseClass):
  success: bool
  datasets: List[AggDatasetObject]
  # anything else you feel should get passed here
```

Then the outer structure of the response objects stays consistent even when nothing is found and the matched datasets are still in there.

tests/test_query.py
surchs: OK, I understand. But even with the old test you aren't checking whether the result is a list, but whether it's an empty list. So before the test read like: 

> "get status 200 and a non-empty result list".

Now the test reads 

> "get status 200 and also make sure that none of the objects in the list have `0` as their `num_matching_subjects` attribute". 

But I think there isn't really a way for that second statement to ever be wrong (if the API runs at all). You mock out the `get` and make it return a list with a known content. And the API essentially takes this list and without doing anything else returns it to your test client. So in other words: what would I learn from this test failing?

rmanaem: > OK, I understand. But even with the old test you aren't checking whether the result is a list, but whether it's an empty list

Yea let me correct myself currently, the result is going to be a non-empty list either way so the old assertion doesn't differentiate between an empty response and a non-empty one.

> But I think there isn't really a way for that second statement to ever be wrong (if the API runs at all). You mock out the get and make it return a list with a known content. And the API essentially takes this list and without doing anything else returns it to your test client. So in other words: what would I learn from this test failing?

I see your point but I feel like that's the case with (tests from the) previous implementation. We always mock the `get` function and assert over the return from the mock and as you mentioned, the API essentially takes that and returns it to the test client without doing anything so there isn't really a way for the second statement to ever be wrong (as long as the API runs with no issue).

I look at the second assertion statement as a way for us to document the expected behavior of the API and our decisions regarding implementation e.g. when we have a valid but not available image modality in the query of the API we decided we would let it pass and return an empty list which in the current implementation would be a list with a dictionary pointing to 0 matched subjects and its test will reflect that.

If I'm not mistaken @alyssadai and I discussed whether it makes sense to have a second assert statement back when we were first setting up tests. I think we can circle back and discuss it in #48.

surchs: OK, makes sense. I thought before we doing some parsing on top of the mocked get return so at least that was being tested. We shouldn't hold up this PR over that question, let's discuss the tests somewhere else and then make a decision.

app/api/crud.py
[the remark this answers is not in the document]
alyssadai: > I think that last rule is a bit confusing. If you give me an empty ghost-dataset if there are no matches in ANY datasets, why don't you just do this for ALL datasets, i.e. always return a list of all datasets in the graph and only set the num_matches to zero for the ones without matches.

Okay, this is a good point.

It sounds then like the clearer options would be to either 1) stick with the original no-matches response of an empty list `[]`, or 2) have it return something like:
```
"success": False
"datasets": []
```
where, when there _are_ matching subjects, `success` would evaluate to `True` and `datasets` would be populated with the dataset-specific dicts as normal.

I think probably then, option 1 (empty list) is still simpler - I don't think wrapping the aggregate dataset objects in another response model necessarily adds anything or makes our lives easier at the level of the frontend (to me `"success": True` is implied by a 200 status code anyways).

@rmanaem I know you were also considering the `[]` response initially, do you have any objections to going back to that behaviour? 😛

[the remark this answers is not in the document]
rmanaem: No problem here 😄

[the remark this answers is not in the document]
alyssadai: Sounds good. I can commit those changes :)

rmanaem:
I think we can revert all tests back to previous implementation now that we decided to go with`[]` in case of an empty response. What do you think @alyssadai @surchs?

tests/test_query.py
```diff
@@ -360,6 +347,9 @@ async def mock_get(
     response = test_app.get(
         f"/query/?image_modal={valid_unavailable_image_modal}"
     )
+
+    print(response.json())
```
surchs, line 351: is that leftover from debuggin?

surchs (approved):
Looks great @rmanaem and @alyssadai! 

There seems to be a `print` statement leftover in the tests. Otherwise good to go from my end. We can talk about how we want to test the mocked stuff in a separate issue. 

Nice job! :tada:

PR #56, opened 2023-01-16
Added optional type annotation to arguments of create_query function in utility.py module

### LINKED ISSUE #53: Update type annotations for optional params in utility.create_query
The [create_query function parameters](https://github.com/neurobagel/bagelapi/blob/main/app/api/utility.py#L39-L45) are all optional with a default value of `None`. Given this combo we could probably make this more in line with Pydantic standards by using the `Optional` type here, as described in: https://docs.python.org/3/library/typing.html#typing.Optional

rmanaem:
Closes #53

### DIFF
--- a/app/api/utility.py
+++ b/app/api/utility.py
@@ -1,6 +1,7 @@
 """Constants for Stardog graph connection and utility functions for writing the SPARQL query."""
 
 from collections import namedtuple
+from typing import Optional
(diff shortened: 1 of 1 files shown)


alyssadai (approved):
Looks great @rmanaem! Thanks for all your work on this PR

PR #59, opened 2023-01-19
Update test workflow to run on push

### LINKED ISSUE #58: Modify github actions workflow to trigger on both push and pull
Will ensure that code coverage computation by Coveralls happens both when a PR is opened _and_ when it is merged (so that the `main` branch coverage is always up to date).

With this change, the `workflow_dispatch` event trigger can probably be dropped.

alyssadai:
`workflow_dispatch` trigger event replaced with `push`.

Closes #58

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -1,6 +1,6 @@
 name: Test
 
-on: [workflow_dispatch, pull_request]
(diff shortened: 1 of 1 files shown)


surchs (approved):
Neat, thanks @alyssadai !

PR #61, opened 2023-01-21
Implemented subject file paths feature and updated test workflow file

### LINKED ISSUE #35: Create queries that only return aggregate information and access information
When I run a cohort query through the API, the results should not contain any demographic metadata (like age, sex, etc) but instead should contain:

- [x] #49
- [x] #50
- [x] #51
- [x] #52

### LINKED ISSUE #51: Return FS paths for subjects that match a query
To do:

- [x] #54
- [x] Query template should be updated to return paths at the **session-level** regardless of the total number of unique matching subjects
- [x] Update `AggDatasetResponse` model
- [x] Update `get` function in `crud.py` module
- [x] ~Test that corresponding field in response body (e.g., `subject_paths`) returns a list~
- [x] Update `test_data` in `test_query.py` module

### LINKED ISSUE #58: Modify github actions workflow to trigger on both push and pull

rmanaem:
Closes #51 
Closes #35 
Closes #58


app/api/crud.py
```diff
@@ -87,7 +87,9 @@ async def get(
                 AggDatasetResponse(
                     dataset=dataset,
                     dataset_name=dataset_name,
-                    num_matching_subjects=group.shape[0],
+                    num_matching_subjects=group["sub_id"].unique().shape[0],
+                    subject_file_paths=list(group["file_path"]),
+                    image_modals=list(group["image_modal"].unique()),
```
alyssadai, line 92: I like having the info about the represented image modalities per dataset, but I think it could be a bit clearer either on the API side or whenever you implement it on the frontend side. For instance, right now a value of
```
"image_modals": ["http://purl.org/nidash/nidm#T1Weighted", "http://purl.org/nidash/nidm#T2Weighted"]
```
for a no-filters query means that at least one subject in each returned dataset has images of each listed modality. I think this output becomes unintuitive to interpret for datasets where not everyone has the same modalities available (e.g., if 100/100 subjects have T1 images, but only 50/100 have T2 images), or for queries where you're filtering by a certain modality. e.g., If a user is just looking for subjects with T1 data, the resulting `image_modals` could still be the same as the above which means that all subjects returned have T1 data, and _at least one_ of those subjects also has T2 data. But I'm not sure this would be immediately clear from the `image_modals` list itself.

I realize that currently the test data provides only datasets where everyone has the same modalities acquired, so this could definitely be a discussion point to revisit.

Some ideas to make this output more intuitive:
- Could we combine the aggregate image modality info with the number of matching subjects? To show the _# subjects per modality_ for a given query.
- We could just document clearly how to interpret the `image_modals` list (i.e., what multiple modalities means when you queried for only one, or the fact that it doesn't necessarily apply to all matching subjects), either in the frontend or in user-facing documentation.

What do you think?

.github/workflows/test.yaml
```diff
@@ -1,6 +1,11 @@
 name: Test
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+  workflow_dispatch:
```
alyssadai, line 8: New implementation looks good :)

alyssadai (commented):
Looks great @rmanaem! 👍 

I have only one question/comment related to the new aggregate image modality info in the query response, which could also be moved to a separate discussion-type issue if preferred.

app/api/utility.py
```diff
@@ -116,10 +116,10 @@ def create_query(
     query_template = f"""
     {DEFAULT_CONTEXT}
 
-    SELECT ?dataset ?dataset_name ?sub_id
+    SELECT ?dataset ?dataset_name ?sub_id ?file_path ?image_modal
     WHERE {{
     SELECT DISTINCT ?dataset ?dataset_name ?subject ?sub_id ?age ?sex
-    ?diagnosis ?image_modal ?num_sessions ?assessment
+    ?diagnosis ?num_sessions ?assessment ?image_modal ?file_path
```
surchs, line 122: Two comments here:

1. I would spell out `?image_modality` here for clarity. Doesn't add that much typing, and makes it just that little bit clearer we aren't talking about a modal
2. We're now re-using these variable names in other parts of the app (e.g. in `crud.py` to group the results and compute numbers). I think we should add an issue to refactor the SPARQL variable names into constants so we can import and reuse them from there and not have to re-type them correctly everytime.

tests/test_query.py
```diff
@@ -15,11 +15,40 @@ def test_data():
             "dataset": "http://neurobagel.org/vocab/qpn",
             "dataset_name": "QPN",
             "num_matching_subjects": 50,
+            "subject_file_paths": [
+                "/my/happy/path/sub-0653/to/session-02",
+                "/my/happy/path/sub-1063/to/session-01",
```
surchs, line 20: How do you pick the number of paths here? The `num_matching_subjects` says 50. Not suggesting you put 50 paths here, rather that you adjust `num_matching_subjects` down to a correct value. But the numbers should match up. I would go for something like 4 paths max to keep this readable. The minimum amount of data that passes the tests.

tests/test_query.py
```diff
@@ -29,14 +58,14 @@ def mock_successful_get(test_data):
...
-        age_min,
-        age_max,
+        min_age,
+        max_age,
         sex,
         diagnosis,
         is_control,
         min_num_sessions,
-        image_modal,
         assessment,
```
surchs, line 67: Why is this now moved up to here? No issue with re-ordering arguments. But there should be a reason, or at least we should agree on one way so we don't reshuffle this every couple of PRs.

tests/test_query.py
```diff
@@ -258,6 +287,28 @@ def test_get_invalid_min_num_sessions(
     response.status_code = 422
 
 
+def test_get_valid_assessment(test_app, mock_successful_get, monkeypatch):
```
surchs, line 290: Same here: nothing wrong with moving things around. But there should be a reason so we don't move it again next week. So in the future let's make an issue for this, discuss the idea of reordering, and then make a separate PR for the reordering so it doesn't get mixed in with unrelated functional changes.

app/api/crud.py
rmanaem: Yea I agree with you, I think we can document it for the first release and revisit it after.

surchs (approved):
Thanks @rmanaem and @alyssadai for this PR! Looks good, I did leave a comment in `crud.py` where I think things could be a bit clearer, but with that it'd be good to go.

A more general comment: Let's keep PRs thematically grouped. That is, if there is a bigger code cleanup or renaming that needs to also happen, do that in a separate PR. 1) it makes the PR easier to understand and review if everything is related, and 2) it gives us a place to discuss the refactor without mixing discussions. 

And if the PR does more than one thing like this one, do add a short sentence as to what the PR does. Linking the issues helps, but ideally the context in the PR is enough to get it.

.github/workflows/test.yaml
rmanaem: Thanks!

app/api/crud.py
surchs: I agree with both of you :) ! @alyssadai do you mind spinning your comment off into its own issue [like so](https://docs.github.com/en/issues/tracking-your-work-with-issues/creating-an-issue#creating-an-issue-from-a-comment) and then we can continue discuss things there?

tests/test_query.py
rmanaem: The `assessment` is moved above `image_modal` and below `min_num_sessions` to follow the filtering order inside `create_query`. Since we don't have inline comments inside different modules I thought having consistent order in which parameters/arguments appear across different modules can help with readability especially for a module with a lot of code e.g., `test_query`

app/api/utility.py
surchs: Ah I see. Well not very likely to cause confusion, but modal does have a separate meaning in a UI context https://en.wikipedia.org/wiki/Modal_window

I mostly like the clarity though. But maybe we discuss this separately because it's already established

rmanaem: Yea let's do it separately

tests/test_query.py
surchs: Ah OK, makes sense to me. Thanks for the explanation

app/api/models.py
```diff
@@ -10,31 +10,31 @@
...
-    age_min: float = Query(default=None, ge=0)
-    age_max: float = Query(default=None, ge=0)
+    min_age: float = Query(default=None, ge=0)
+    max_age: float = Query(default=None, ge=0)
     sex: Literal["male", "female", "other"] = None
     diagnosis: constr(regex=r"^[a-zA-Z]+[:]\S+$") = None
     is_control: bool = None
     min_num_sessions: int = Query(default=None, ge=1)
-    image_modal: constr(regex=r"^[a-zA-Z]+[:]\S+$") = None
     assessment: constr(regex=r"^[a-zA-Z]+[:]\S+$") = None
```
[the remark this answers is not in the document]
rmanaem: See #64

PR #71, opened 2023-02-01
Change variable storing graph DB IP address to an environment variable

### LINKED ISSUE #68: Change DOG_ROOT in utility.py to environment variable
Local deployments of API need to point to http://206.12.99.17, but deployment on cloud instance should talk to localhost since our Stardog db is hosted on the same instance.

alyssadai:
`DOG_ROOT` in utility.py will now use `DOG_ROOT` in the user's environment if it's set, or otherwise have a default value of "206.12.99.17".

Closes #68

### DIFF
--- a/app/api/utility.py
+++ b/app/api/utility.py
@@ -1,13 +1,14 @@
 """Constants for Stardog graph connection and utility functions for writing the SPARQL query."""
 
+import os
 from collections import namedtuple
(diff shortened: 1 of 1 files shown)


surchs (approved):
Very nice, good call with the fallback to the public ip 🎉

PR #77, opened 2023-02-06
Implement error catching for `httpx.ConnectTimeout` when not connected to McGill network

### LINKED ISSUE #70: Implement check for requests sent from a non McGill-IP address
For local/dev launches of the API, the docs can be reached but queries from a non-McGill network IP (e.g., if you're not connected to the VPN) will result in an internal server error (`httpx.ConnectTimeout` behind the scenes). To make this more transparent, we could potentially add an initial [check for the client's IP address](https://fastapi.tiangolo.com/advanced/using-request-directly/#details-about-the-request-object) and provide an informative error if it's not within the McGill IP blocks we've granted access to on OpenStack.

alyssadai:
API now returns an informative error message when a `ConnectTimeout` error is encountered (i.e., when attempting to submit a query to the graph while not connected to the McGill network).

Closes #70

### DIFF
--- a/app/api/crud.py
+++ b/app/api/crud.py
@@ -3,7 +3,7 @@ import os
 
 import httpx
 import pandas as pd
-from fastapi import HTTPException
+from fastapi import HTTPException, status
 
 from . import utility as util
 from .models import AggDatasetResponse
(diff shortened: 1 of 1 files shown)


surchs (approved):
Very cool! Thanks for addressing this. Good to go 🎉

PR #81, opened 2023-02-10
Modified `query_template` in utility.py

### LINKED ISSUE #44: Add bg:isSubjectGroup to SPARQL query pattern in create_query function
The SPARQL query template currently doesn't include a triple pattern for `bg:isSubjectGroup` (although a filter using the variable name `subject_group` is added when `is_control=True`) as this relation does not currently exist in the graph.

Once subject group has been added to the neurobagel datamodel, update the query template as follows in utility.py:

```
...
?subject a bg:Subject;
            bg:label ?sub_id;
            bg:age ?age;
            bg:sex ?sex;
            bg:diagnosis ?diagnosis;
            bg:isSubjectGroup ?subject_group;
            bg:hasSession/bg:hasAcquisition/bg:hasContrastType ?image_modal.
...
```

### LINKED ISSUE #72: Modify `query_template` in utility.py module
https://github.com/neurobagel/api/blob/10d86b94d2c880105d71019cd1647639750b49e4/app/api/utility.py#L129-L137

- [x] Change subject properties in `query_template` to optional so that if a property is missing for a subject, the query still returns results.

- [x] Remove resulting `NaN` file paths in query result when forming aggregate response

rmanaem:
Main changes:
- Added `bg:isSubjectGroup` to `query_template`, Closes #44 
- Wrapped age, sex, diagnosis, subject group, and assessment with sparql optional pattern, Closes #72

### DIFF
--- a/app/api/crud.py
+++ b/app/api/crud.py
@@ -94,7 +94,7 @@ async def get(
                     dataset=dataset,
                     dataset_name=dataset_name,
                     num_matching_subjects=group["sub_id"].nunique(),
-                    subject_file_paths=list(group["file_path"]),
+                    subject_file_paths=list(group["file_path"].dropna()),
                     image_modals=list(group["image_modal"].unique()),
                 )
             )
--- a/app/api/utility.py
+++ b/app/api/utility.py
@@ -131,8 +131,9 @@ def create_query(
             bg:hasSession ?session;
             bg:hasSession/bg:hasAcquisition/bg:hasContrastType ?image_modal.
 
-    ?session bg:filePath ?file_path.
-
+    OPTIONAL {{
+        ?session bg:filePath ?file_path.
+    }}
     OPTIONAL {{
         ?subject bg:age ?age.
     }}
--- a/app/api/utility.py
+++ b/app/api/utility.py
@@ -120,7 +120,7 @@ def create_query(
     SELECT ?dataset ?dataset_name ?sub_id ?file_path ?image_modal
     WHERE {{
     SELECT DISTINCT ?dataset ?dataset_name ?subject ?sub_id ?age ?sex
-    ?diagnosis ?num_sessions ?assessment ?image_modal ?file_path
+    ?diagnosis ?subject_group ?num_sessions ?assessment ?image_modal ?file_path
     WHERE {{
     ?dataset a bg:Dataset;
              bg:label ?dataset_name;
@@ -131,6 +131,7 @@ def create_query(
             bg:age ?age;
             bg:sex ?sex;
             bg:diagnosis ?diagnosis;
+            bg:isSubjectGroup ?subject_group;
             bg:hasSession ?session;
             bg:assessment ?assessment;
             bg:hasSession/bg:hasAcquisition/bg:hasContrastType ?image_modal.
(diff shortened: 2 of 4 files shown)


alyssadai (commented):
Thanks @rmanaem! Is there a reason we're keeping `bg:hasSession` as non-optional here for now? Does it break the below `SELECT` block if we also wrap it in an `OPTIONAL` group?

rmanaem:
> Is there a reason we're keeping bg:hasSession as non-optional here for now? Does it break the below SELECT block if we also wrap it in an OPTIONAL group?

I'd refer this to @surchs

surchs:
Aha, excellent question @alyssadai. I think you are right, `hasSession` should be optional because it is optional in our [data model](https://github.com/neurobagel/bagelbids/blob/main/bagelbids/models.py#L43). Making the `hasSession` optional is not a problem from the SPARQL perspective, I just checked. But there are some questions about how we want to use the results. 

At the moment, we expect to return a session file-path for every subject that meets the the cohort criteria because (at least in the API) we expect every subject to have at least one session. My view on this at the moment is:
- let's make the file_path optional. You only get a file_path if you have imaging data
- let's not make the session optional any more. If all your subjects have only one session, BIDS lets you decide if you want to create a session layer at all - but: you can. I would say: let's always have a session for subjects. As soon as we add the phenotypic attributes to the session, we'll need the session anyway. And it makes our lives easier if we don't have to handle has-session and has-not-session paths for all eternity.

It so happens that I fortuitously failed to create a file_path attribute for all `"Queensland Twin IMaging (QTIM)"` subjects and sessions  in the test graph instance. So we can actually start experimenting with this case now.

alyssadai:
Thanks for the explanation @surchs. Your second bullet point makes sense to me (and I like always having a session for subjects), but just to clarify the first
> let's make the file_path optional. You only get a file_path if you have imaging data

What do you mean by optional? Would participants without imaging data just not be annotated with paths in the graph in the first place? So, from the API aggregate query results perspective (assuming there are participants in the graph who have only _non-imaging_ data), you would still get a list of paths for your query, but the # participants with paths in that list may not be the same as the total # matching participants for the given query (?)

surchs:
>  you would still get a list of paths for your query, but the # participants with paths in that list may not be the same as the total # matching participants for the given query

Thanks, that's a good point. I think we haven't really discussed what the decision that "you can exist without neuroimaging data" means for the API userflow - and to a lesser extent also the query tool. My view is that for this PR / release, your suggestion of returning a list of paths with possibly fewer entries than the stated number of subject*sessions is fine. 

The main purpose of the paths at the moment is so that a user can get to the binary imaging files if they want to. We don't really have something similar to provide to the user yet for a subject without imaging data, but we still want to mention that these subjects exist. 

That said, we should revisit the question of "what can we tell a user about a pheno-only subject" quickly. I made https://github.com/neurobagel/api/issues/84 to track this.

With that, I'd say we can go ahead here and make the file_path attribute optional. LMK if you want to look at this together @rmanaem

alyssadai (approved):
Looks great @rmanaem, thanks for catching + addressing this quirk with the SPARQL results!

I've added a commit to drop the NaN file paths in the aggregate query response. With this change the `ORJSONResponse` class should no longer be needed, but I think it's fine if we keep it in for now in case we decide to revert to returning NaN values for attributes.

PR #82, opened 2023-02-10
bump isort to v5.12.0

### LINKED ISSUE #80: bump isort pre-commit to v15.12.x
As described in this stackoverflow post: https://stackoverflow.com/questions/75269700/pre-commit-fails-to-install-isort-5-11-4-with-error-runtimeerror-the-poetry-co there is a problem with isort < v15.12 having an unpinned dependency on poetry and being incompatible with the most recent poetry version. if you run pre-commit with the pinned isort version (https://github.com/neurobagel/api/blob/main/.pre-commit-config.yaml#L26) on a new/updated environment, the pre-commit will crash.

Simple fix is to pin v15.12.0 for isort in the pre-commit yaml file

alyssadai:
Closes #80

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -23,7 +23,7 @@ repos:
       - id: check-yaml
 
   - repo: https://github.com/pycqa/isort
-    rev: 5.10.0
+    rev: 5.12.0
     hooks:
       - id: isort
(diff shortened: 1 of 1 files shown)


alyssadai:
Merging without review as all checks are passing

PR #90, opened 2023-04-03
Add docker-compose specification and instructions

### LINKED ISSUE #78: Deploy API in Docker container on cloud server
- [x] Ensure that the dockerized API can connect to the graph locally
  - the connection should happen inside the server. So either directly from docker API to graph on the same machine or
  - within a shared docker network between docker API and docker graph
- [x] #86
- [x] #87

### LINKED ISSUE #86: Add API and graph to docker-compose specification
- [x] create docker-compose.yml file containing two services, `api` and `graph`
- [x] update README with instructions for how to spin up docker-compose

alyssadai:
Other changes:
The `USER` variable has been renamed everywhere to `USERNAME` to try to avoid name conflicts on Linux systems where `USER` is automatically set in the shell. This is more of a precaution; a note has also been added in the README which recommends that the user verifies that the values of the credentials pair is correct.

Closes #86 
Closes #78

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,4 +8,6 @@ RUN pip install --no-cache-dir --upgrade -r /usr/src/app/requirements.txt
 
 COPY ./app /usr/src/app
 
-ENTRYPOINT ["uvicorn", "app.main:app", "--proxy-headers", "--host", "0.0.0.0", "--port", "8000"]
+ENTRYPOINT ["uvicorn", "app.main:app", "--proxy-headers", "--host", "0.0.0.0"]
+
+CMD ["--port", "8000"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,9 +7,3 @@ COPY ./requirements.txt /usr/src/app/requirements.txt
 RUN pip install --no-cache-dir --upgrade -r /usr/src/app/requirements.txt
 
 COPY ./app /usr/src/app
-
-ENTRYPOINT ["uvicorn", "app.main:app", "--proxy-headers", "--host", "0.0.0.0"]
-
-CMD ["--port", "8000"]
-
-
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,3 +7,5 @@ COPY ./requirements.txt /usr/src/app/requirements.txt
 RUN pip install --no-cache-dir --upgrade -r /usr/src/app/requirements.txt
 
 COPY ./app /usr/src/app
+
+ENTRYPOINT ["uvicorn", "app.main:app", "--proxy-headers", "--host", "0.0.0.0", "--port", "8000"]
--- a/README.md
+++ b/README.md
@@ -124,6 +124,9 @@ INFO:     Application startup complete.
 ```
 You can verify the API is running once you receive info messages similar to the above in your terminal.
 
+### Troubleshooting
+If you get a 401 response to your API request with an `"Unauthorized: "` error message, your `USERNAME` and `PASSWORD` pair may be incorrect. Verify that these environment variables have been exported and/or have the correct values.
+
 ## Testing
 
 Neurobagel API utilizes [Pytest](https://docs.pytest.org/en/7.2.x/) framework for testing.
--- a/README.md
+++ b/README.md
@@ -39,16 +39,21 @@ NOTE: Currently, to access the API, you must be connected to the McGill network.
 ## Local installation
 
 ### Set the environment variables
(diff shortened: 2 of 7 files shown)


README.md
```diff
@@ -37,45 +39,62 @@ NOTE: Currently, to access the API, you must be connected to the McGill network.
...
+- `API_TAG`: Tag for API Docker image (default: `latest`)
+- `DOG_ROOT`: container name or IP address for the graph database (default: `graph`)
+- `STARDOG_TAG`: Tag for Stardog Docker image (default: `7.7.3-java11-preview`)
+- `STARDOG_ROOT`: Path to directory on host machine containing a Stardog license file (default: `~/stardog-home`)
+
+Note: To avoid conflicts related to [Docker's environment variable precedence](https://docs.docker.com/compose/environment-variables/envvars-precedence/), ensure that any variables defined in your `.env` file are not already set in your current shell environment with **different** values.
+
+Then spin up the containers using Docker Compose:
+```bash
+docker compose up -d
```
surchs, line 76: you could add a note that you need to run this command in the same dir as the `docker-compose.yml` file, or you need to specify the path via `-f` (ask me how I know :) )

README.md
```diff
@@ -37,45 +39,62 @@ NOTE: Currently, to access the API, you must be connected to the McGill network.
...
+source .env # set your environment variables
 docker build -t <image_name> .
-docker run -d --name api -p 8000:8000 --env USER --env PASSWORD neurobagel/api
+docker run -d --name api -p 8000:8000 --env USERNAME --env PASSWORD neurobagel/api
 ```
-For either option, if you wish to also set `DOG_ROOT`, make sure to pass it to the container in the `docker run` command using the `--env` flag.
+
+For Options 2 or 3, if you wish to also set `DOG_ROOT`, make sure to pass it to the container in the `docker run` command using the `--env` flag.
 
 NOTE: In case you're connecting to the McGill network via VPN and you started the container before connecting to the VPN, make sure to configure your VPN client to allow local (LAN) access when using the VPN.
```
surchs, line 96: Two things I would add here:

1) I have just started everything for the first time (like you said): now what do I do? is there some link I can click that should now open?
2) It'd be nice to add a line on: "OK, if you run this all for the first time, you will need to create a graph database to be queried (and put data in it). The easiest way is probably to use Stardog-Studio: https://docs.stardog.com/stardog-applications/dockerized_access, point it at `localhost:5820` (unless you changed that) and login with `admin, admin`. Good luck with that"

surchs (approved):
OK, this an absolutely amazing PR @rmanaem and @alyssadai :tada:! I really love the detailed README. What a joy to read and follow. I tried out the steps, it all works well. Big fan of the hand-holding with setting up docker and docker-compose.

I left some :cherries: suggestions for the README. The one for the API Dockerfile is a bit more of a question - but take a look if you agree, and then good to go :cook:.

surchs:
Oh, before I forget: two things I realized as I was reading the PR and I made issues for them:

1. We still have a hard-coded variable in the API for the name of the graph database that we send requests to: https://github.com/neurobagel/api/blob/e996fe213219f3d933dd4c579dcec60023a588ee/app/api/utility.py#L9 this should be changed into a configurable (environment variable or runtime option).  #91 
2. We need to discuss / decide how we want people to create a stardog user, a new stardog database, and then fill this database with data after they have run `docker compose up`. Ideally we can provide a fully automated way to spin all of this up so that they only have to push some data into the graph before they can get started #92

Dockerfile
```diff
@@ -8,8 +8,4 @@ RUN pip install --no-cache-dir --upgrade -r /usr/src/app/requirements.txt
 
 COPY ./app /usr/src/app
 
-ENTRYPOINT ["uvicorn", "app.main:app", "--proxy-headers", "--host", "0.0.0.0"]
-
-CMD ["--port", "8000"]
-
-
+ENTRYPOINT ["uvicorn", "app.main:app", "--proxy-headers", "--host", "0.0.0.0", "--port", "8000"]
```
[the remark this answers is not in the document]
alyssadai: Mostly for simplicity, as to override `CMD` in the Dockerfile we would have to add a `command:` directive in `docker-compose.yml`, and then I was thinking we would have to code for extra environment variables for the ports (at this point it almost makes more sense for the user to create a new `yml`). 

That said, upon reflection, since the Docker image for the API will likely be pulled from Dockerhub in most cases, I think it makes sense to keep this configurability in and trust that if a user really wanted to customize all the ports (for the API or graph), they could modify the docker-compose file itself/create one that suits their needs (without having to change anything in the Dockerfile). So I've reversed this change without adding any additional environment variable substitutions in `docker-compose.yml`.

PR #93, opened 2023-04-11
Make graph database name configurable as an environment variable

### LINKED ISSUE #91: Make graph name configurable
Right now the name of the graph we send requests to is hard-coded: https://github.com/neurobagel/api/blob/e996fe213219f3d933dd4c579dcec60023a588ee/app/api/utility.py#L9

Instead it should be configurable. Options could be
- environment var (like username and password)
- a runtime configurable flag or variable (that also works with docker)

alyssadai:
- Default Stardog image tag also updated to reflect our current license.

Closes #91


surchs (approved):
Awesome, thanks for doing this so quickly @alyssadai. Good to go :cook:

PR #99, opened 2023-04-24
Return (or not) subject-level attributes in query results based on env var

### LINKED ISSUE #96: Create option to return subject level metadata
Our API currently wraps the SPARQL query in a query that aggregates results. For truly open deployments (like Open Neuro), we would like to not aggregate the results. But it is important that we don't accidentally turn on the "return everything" option if we didn't mean to / when we really expected to only return aggregate data.

We should:

- [x] provide an option during API launch to OPT IN to returning everything
  - [x]  can add new environment variable `RETURN_AGG` where default is `false`
- [ ] test that not opting into this mode does only return aggregate data
- [x] update the query template to return the full results
- [x] update the processing of the results returned from the Graph so it can handle the sample-level results
  - [x] keep cohort info: dataset id, dataset name, # matching subjects

For reference: this was the PR where we added the aggregate results https://github.com/neurobagel/api/commit/94243b809da6bb5a7e376d45fd960231c255b176#diff-c996069b15819e38d0de4c23df3a45ca9b56c4315a80d8b05622916543011829

alyssadai:
Closes #96

### DIFF
--- a/app/api/utility.py
+++ b/app/api/utility.py
@@ -119,8 +119,6 @@ def create_query(
     query_template = f"""
     {DEFAULT_CONTEXT}
 
-    SELECT ?dataset ?dataset_name ?sub_id ?file_path ?image_modal
-    WHERE {{
     SELECT DISTINCT ?dataset ?dataset_name ?subject ?sub_id ?age ?sex
     ?diagnosis ?subject_group ?num_sessions ?assessment ?image_modal ?file_path
     WHERE {{
@@ -165,7 +163,6 @@ def create_query(
 
     {subject_level_filters}
 }}
-}} GROUP BY ?dataset ?dataset_name ?sub_id ?file_path ?image_modal
 """
 
     return query_template
--- a/app/api/utility.py
+++ b/app/api/utility.py
@@ -41,6 +41,7 @@ IS_CONTROL_TERM = "http://purl.obolibrary.org/obo/NCIT_C94342"
 
 
 def create_query(
+    return_agg: bool,
     age: Optional[tuple] = (None, None),
     sex: Optional[str] = None,
     diagnosis: Optional[str] = None,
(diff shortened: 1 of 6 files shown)


app/api/utility.py
```diff
@@ -165,7 +167,7 @@ def create_query(
 
     {subject_level_filters}
 }}
-}} GROUP BY ?dataset ?dataset_name ?sub_id ?file_path ?image_modal
+{"}}}} GROUP BY ?dataset ?dataset_name ?sub_id ?file_path ?image_modal" if return_agg else ""}
```
surchs, line 170: Where is this `{"}` coming from here?

app/api/utility.py
```diff
@@ -118,9 +121,8 @@ def create_query(
 
     query_template = f"""
     {DEFAULT_CONTEXT}
+    {"SELECT ?dataset ?dataset_name ?sub_id ?file_path ?image_modal WHERE {{{{" if return_agg else ""}
```
surchs, line 124: I think this is a bit mind-bendy. I believe a slightly clearer version would be to build the template in 2 steps:

1. Build the inner query that returns the subject level data
2. IF `return_agg`, then wrap it in the outer query (prepend with the SELECT, append the GROUPBY)
3. ELSE: just return the subject-level template.

In the interest of time, I would push that to the next PR with an issue, but I think we should keep these two steps separated and not use inline if-statements in the f-string.

app/api/utility.py
surchs: Ah ok, that's the counterpart to the if-statement in the f-string above.

surchs (commented):
Looks good @alyssadai, thanks for getting to this so quickly! I left a comment on the query-string template. I think it's all good for now, but we should revisit it in a future PR to make it a bit clearer. Could you make an issue to track that, and then merge! :cook:

surchs (approved):
Aha, meant to approve this. Clicked too fast. Approved with above comments :cook:

PR #101, opened 2023-04-25
Refactor generation of query template

### LINKED ISSUE #100: Refactor query template construction
Instead of using inline if-statements for results aggregation in the query template f-string,

1. Build the inner query that returns the subject level data
2. IF `return_agg`, then wrap it in the outer query (prepend with the SELECT, append the GROUPBY)
3. ELSE: just return the subject-level template.

_Originally posted by @surchs in https://github.com/neurobagel/api/pull/99#discussion_r1175808416_

alyssadai:
Closes #100

### DIFF
--- a/app/api/utility.py
+++ b/app/api/utility.py
@@ -119,55 +119,56 @@ def create_query(
             "\n" + f"FILTER (?{IMAGE_MODAL.var} = {image_modal})."
         )
 
-    query_template = f"""
-    {DEFAULT_CONTEXT}
-    {"SELECT ?dataset ?dataset_name ?sub_id ?file_path ?image_modal WHERE {{{{" if return_agg else ""}
+    all_attributes_query = f"""
+        SELECT DISTINCT ?dataset ?dataset_name ?subject ?sub_id ?age ?sex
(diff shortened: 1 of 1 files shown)


app/api/utility.py
```diff
@@ -119,55 +119,56 @@ def create_query(
...
+            }}
+            {subject_level_filters}
+        }}
+    """
 
-    return query_template
+    if not return_agg:
+        return "\n".join([DEFAULT_CONTEXT, all_attributes_query])
+
+    return "\n".join(
```
surchs, line 167: :cherries: I still find this a bit mentally-difficult. So far, we have made a query string. If this flag is not set, we will wrap it in the aggregating query string. So we could just say 

```python
query_string = f"SELECT ?dataset ?dataset_name ?sub_id ?file_path ?image_modal WHERE {{\n{query}\n
"}} GROUP BY ?dataset ?dataset_name ?sub_id ?file_path ?image_modal"
```

and then just return `query_string` with the DEFAULT_CONTEXT.

surchs (approved):
Looks good @alyssadai. I left one ask and one :cherries: comment. With the added comment I think this is good to go :cook:

PR #124, opened 2023-05-25
[FIX] Update names + storage of env variables to avoid conflicts + for ease of access

### LINKED ISSUE #114: [FIX] Avoid using shell variable names that are likely to clash

- [X] I have searched the existing issues

### New feature

The current README instructions expect variables in the shell to be called `USERNAME` and `PASSWORD` for configuring the `docker compose` environment. That's not a good idea, because these shell variable names are likely to be already set by some other configuration, leading to the need to first unset them (as outlined as well in the readme).

For example, in the z-shell, `USERNAME` is set to `$USER` by default (see https://linux.die.net/man/1/zshparam).

A better solution would be to use variable names that are less likely to clash in this way (or find some other way to override the shell primacy for these variables), such as by appending a unique prefix `NB_`.

E.g.:

`NB_GRAPH_USERNAME` and `NB_GRAPH_PASSWORD`

Since the username and password may be graph-specific, having `GRAPH` in the variable name can help to further minimize confusion about the purpose of the credentials.

### Unclear documentation

When fixed, update the README as well as the corresponding warning in the documentation: https://www.neurobagel.org/documentation/start/#launch-the-api-and-graph-stack

### Other steps to implement
- to avoid having to change names across modules (incl. any tests) when an environment variable accessed from within the API is renamed, can transform these into `namedtuple` variables with "name" and "val" fields
- when env vars are updated, update API tests

### LINKED ISSUE #115: [BUG] GRAPH_DB defaults to empty string, not `test_data`

- [X] I have searched the existing issues

### Expected Behavior

If `GRAPH_DB` is not set in the `.env`, it should default to `test_data` on the API side as per the [README](https://github.com/neurobagel/api/blob/9d618785fd309d8e884f0d440a7964227d97b582/README.md#option-1-use-the-docker-composeyaml-file)

### Current Behavior

When `GRAPH_DB` is unset, `docker compose` sets it to `""` (i.e. empty string) by default so that it exists in the environment.

Below is the output of `docker compose config`:

```
WARN[0000] The "GRAPH_DB" variable is not set. Defaulting to a blank string.
WARN[0000] The "RETURN_AGG" variable is not set. Defaulting to a blank string. 
name: stardog-trial-run
services:
  api:
    environment:
      GRAPH_ADDRESS: graph
      GRAPH_DB: ""
      PASSWORD: admin
      RETURN_AGG: ""
      USERNAME: admin
    image: neurobagel/api:latest
...
```
This is due to the variable being defined in the `docker compose` environment.

As a result, there is an unintended effect where the API looks for the `GRAPH_DB` in the environment, sees that it has been set (just to a `""`) and thus does not use the default name: https://github.com/neurobagel/api/blob/9d618785fd309d8e884f0d440a7964227d97b582/app/api/utility.py#L9

### Environment

docker compose

### Anything else?

This is also happening invisibly for `RETURN_AGG`, the other env variable which is defined in the docker compose environment but does not have a default value in the `docker-compose.yml` (but _does_ on the API side). However, since the API checks equivalence of the `RETURN_AGG` value rather than just fetching it from the env, the behaviour is still as expected even when `RETURN_AGG` is unset:
https://github.com/neurobagel/api/blob/9d618785fd309d8e884f0d440a7964227d97b582/app/api/crud.py#L14

If I don't notice that `GRAPH_DB` defaults to empty,
and just expect the API to send requests to `test_data`,
then I will be a bit confused by the current error code returned from the API,
which is:
```
{"detail":"Not Found: Not Found!"}
```
This should be documented in the README.

### Steps to fix
- [ ] Add default values for `GRAPH_DB` and `RETURN_AGG` in docker-compose.yml

alyssadai:
Closes #114, closes #115

- Current environment variables (variables that can currently be configured in a user/sysadmin shell) have been renamed to have an `NB` prefix to minimize clashes with other shell vars
- Internal storage of environment variables has been changed to use `namedtuples` so that variable names can be changed/accessed more programmatically, and is handled by utility.py to keep things in one place; tests updated
- Table added to README to list and describe all environment variables in one place

## Checklist

- [X] PR has an interpretable title with a prefix (`[ENH]`, `[BUG]`, `[DOC]`, `[INFRA]`, `[MAINT]`)
- [X] PR links to Github issue with mention `Closes #XXXX`
- [X] Tests pass
- [X] Code is properly formatted

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,39 +47,45 @@ git clone https://github.com/neurobagel/api.git
 ```
 
 ### Set the environment variables
-Create a `.env` file in the root of the repository to house the environment variables used by the app. To run the API, at least two environment variables must be set, `USERNAME` and `PASSWORD`.  
+Create a `.env` file in the root of the repository to house the environment variables used by the app. 
+
+To run API requests against a graph, at least two environment variables must be set, `NB_GRAPH_USERNAME` and `NB_GRAPH_PASSWORD`.  
 The contents of a minimal `.env` file:
 ```bash
-USERNAME=someuser
-PASSWORD=somepassword
+NB_GRAPH_USERNAME=someuser
+NB_GRAPH_PASSWORD=somepassword
 ```
-Optional environment variables for the API may also be defined in `.env`:
-- `GRAPH_ADDRESS`: to use a different IP address for the graph database
-- `RETURN_AGG`: whether to return only dataset-level query results (including file paths) and exclude subject-level attributes. Should be one of [true, false], true by default.
 
-To export all the variables in your `.env` file in one step, run the following:
+Below are all the possible Neurobagel environment variables that you can set in `.env`, depending on your mode of installation of the API.
+| Environment variable | Required in .env? | Description                                                                                                                              | Default value            | Relevant installation mode |
+|----------------------|-------------------|------------------------------------------------------------------------------------------------------------------------------------------|--------------------------|----------------------------|
+| `NB_GRAPH_USERNAME`  | Yes               | Username to access Stardog graph database that API will communicate with                                                                 | -                        | Docker, Python             |
+| `NB_GRAPH_PASSWORD`  | Yes               | Password to access Stardog graph database that API will communicate with                                                                 | -                        | Docker, Python             |
+| `NB_GRAPH_ADDRESS`   | No                | IP address for the graph database (or container name, if graph is hosted locally)                                                        | `206.12.99.17` (`graph`) | Docker, Python             |
+| `NB_GRAPH_DB`        | No                | Name of graph database to query                                                                                                          | `test_data`              | Docker, Python             |
+| `NB_RETURN_AGG`      | No                | Whether to return only dataset-level query results (including data locations) and exclude subject-level attributes. One of [true, false] | `true`                   | Docker, Python             |
+| `NB_API_TAG` | No | Tag for API Docker image | `latest` | Docker
+| `STARDOG_TAG` | No | Tag for Stardog Docker image | `8.2.2-java11-preview` | Docker |
+| `STARDOG_ROOT` | No | Path to directory on host machine containing a Stardog license file | `~/stardog-home` | Docker |
+
+NOTE: To avoid conflicts related to [Docker's environment variable precedence](https://docs.docker.com/compose/environment-variables/envvars-precedence/), ensure that any variables defined in your `.env` file are not already set in your current shell environment with **different** values.
+
+To export all the variables defined in your `.env` file in one step, run the following:
 ```bash
 export $(cat .env | xargs)
 ```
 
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -88,19 +94,19 @@ docker compose up -d
 
 #### Option 2: Use the latest image from Docker Hub
 ```bash
-source .env # set your environment variables 
+export $(cat .env | xargs)  # export your environment variables
```
surchs, line 97: interesting. were they not getting picked up as shell variables?

docker-compose.yml
```diff
@@ -2,15 +2,15 @@ version: "3.8"
...
+    image: "neurobagel/api:${NB_API_TAG:-latest}"
     ports:
       - "8000:8000"
     environment:
-      USERNAME: ${USERNAME}
-      PASSWORD: ${PASSWORD}
-      GRAPH_ADDRESS: ${GRAPH_ADDRESS:-graph}
-      GRAPH_DB: ${GRAPH_DB}
-      RETURN_AGG: ${RETURN_AGG}
+      NB_GRAPH_USERNAME: ${NB_GRAPH_USERNAME}
```
surchs, line 9: just for my understanding: is this the place where we can independently decide what the env variables should be called outside docker (i.e. on user machine) and inside docker (i.e. our app)? 

In other words, Could I change this line to
```
HEHE_YOU_WILL_NEVER_KNOW_VAR: ${NB_GRAPH_USERNAME}
```
and be OK as long as that's what I am using internally in my `os.getenv` calls?

surchs (approved):
Thanks for addressing this @alyssadai. All looks good. I have one question about the env-vars and your refactoring of them. I wouldn't hold up the PR for this, but I think it's good to clarify possible issues while this is still fresh in memory.

One thing that I think we do need and should address in a separate issue/PR is some form of a test to prevent a recurrence of "whoops the defaults did things we didn't expect". Could you please open an issue to think of and implement a test that would fail with the old error we encountered.

surchs:
Oh, :cook: of course!

README.md
alyssadai: The `source` command does set the variables from .env in the shell, but doesn't export them. This is probably not absolutely necessary (esp with the new variable prefixes) but more of an extra precaution to avoid conflicts with existing shell variables. 🙂

docker-compose.yml
alyssadai: Yes. In your example `NB_GRAPH_USERNAME` would need to be set in the host env, it gets assigned to the variable `HEHE_YOU_WILL_NEVER_KNOW_VAR`  inside Docker (in the container's env), and we would update the corresponding global variable in utility.py to be
```python
GRAPH_USERNAME = EnvVar(
    "HEHE_YOU_WILL_NEVER_KNOW_VAR", os.environ.get("HEHE_YOU_WILL_NEVER_KNOW_VAR")
)
```

PR #146, opened 2023-08-16
[ENH] Increase configurability of API/graph server parameters

### LINKED ISSUE #118: [ENHANCEMENT] Provide a template .env file
An .env file with all of the necessary variables should be provided with their defaults.

### LINKED ISSUE #122: [ENHANCEMENT] ports should be configurable
Looks like stardog hard-codes its port internally, but you can remap it externally.

This means the API server needs to accept a port # for the stardog instance.

The API server itself should also have a configurable port.

### LINKED ISSUE #140: Increase httpx timeout because our queries are so damn slow
So apparently #139 is breaking because the queries are now too slow for httpx. And once it closes its connection the graphdb has a moment and crashes. 

So what we should do as a (very short term) fix is to set the httpx timeout to some unfathomably large number, let's say 1 minute and then rebuild the image and deploy. See if that helps.

What we should also do shortly after is make the queries less slow.

### LINKED ISSUE #142: Generalize graph ports and docker config for different graph backends
- [x] Make the following configurable as env vars:
  - [x] graph ports and API ports
    - #122 
  - [x] graph Docker image
  - [x] graph home **and** root paths
  - [x] full graph db query endpoint

- [x] Update README

alyssadai:
Closes #118
Closes #122
Closes #140 
Closes #142
Closes https://github.com/neurobagel/project/issues/162

- Make the following configurable as environment variables:
  - host + container ports
  - graph server Docker image
  - graph root paths on host + container
  - full graph database query endpoint
- Update README with new env vars + simplified instructions for using .env file when using building or using API Docker image alone
- Increase timeout for query to minimize risk of occasional timeouts
- Add an annotated template .env file

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:


Dockerfile
```diff
@@ -8,6 +8,4 @@ RUN pip install --no-cache-dir --upgrade -r /usr/src/app/requirements.txt
 
 COPY ./app /usr/src/app
 
-ENTRYPOINT ["uvicorn", "app.main:app", "--proxy-headers", "--host", "0.0.0.0"]
-
-CMD ["--port", "8000"]
\ No newline at end of file
+ENTRYPOINT uvicorn app.main:app --proxy-headers --host 0.0.0.0 --port ${NB_API_PORT:-8000}
```
surchs, line 11: 1. I feel like the removal of `CMD` keeps showing up in some PRs but I don't think we want that, no? [`CMD`](https://docs.docker.com/engine/reference/builder/#cmd) gives me a default I can easily change. [`ENTRYPOINT`](https://docs.docker.com/engine/reference/builder/#entrypoint) gives me a default I shouldn't have to change. Port is probably something we want to have as a CMD default then
2. `${NB_API_PORT:-8000}` this looks cool - but how does it get in here? I would have expected some `ARG` or `ENV` command to set it first? More importantly: what is it supposed to do? Let you change the default port at build time?
3. I'm assuming there is a reason you changed this line from the `exec` form to the `shell` form - is this linked to the appearance of the shell variable and the disappearance of the CMD?

README.md
```diff
@@ -56,60 +56,71 @@ NB_GRAPH_USERNAME=someuser
...
+| `NB_RETURN_AGG`      | No                | Whether to return only dataset-level query results (including data locations) and exclude subject-level attributes. One of [true, false] | `true`                                 | Docker, Python                |
+| `NB_API_TAG`         | No                | Tag for API Docker image                                                                                                                 | `latest`                               | Docker                        |
+| `NB_API_PORT_HOST`   | No                | Port number on the _host machine_ to map the API container port to                                                                       | `8000`                                 | Docker, Python                |
+| `NB_API_PORT`        | No                | Port number to use for the _API container_                                                                                               | `8000`                                 | Docker, Python                |
```
surchs, line 69:
```suggestion
| `NB_API_PORT`        | No                | Port number to expose on the _API container_                                                                                               | `8000`                                 | Docker, Python                |
```
does that fit better?

README.md
```diff
@@ -56,60 +56,71 @@ NB_GRAPH_USERNAME=someuser
...
+| `NB_GRAPH_ROOT_HOST` | No                | Path to directory containing a Stardog license file on the _host machine_                                                                | `~/stardog-home`                       | Docker                        |
+| `NB_GRAPH_ROOT_CONT` | No                | Path to directory for graph databases in the _graph server container_                                                                    | `/var/opt/stardog` *                   | Docker                        |
+| `NB_GRAPH_PORT_HOST` | No                | Port number on the _host machine_ to map the graph server container port to                                                              | `5820`                                 | Docker                        |
+| `NB_GRAPH_PORT`      | No                | Port number used by the _graph server container_                                                                                         | `5820` *                               | Docker                        |
+
+_* These defaults should not be changed if using a Stardog backend._
```
surchs, line 76:
```suggestion
_* These defaults are configured for a Stardog backend - you should not have to change them if you are running a Stardog backend._
```

README.md
```diff
@@ -56,60 +56,71 @@ NB_GRAPH_USERNAME=someuser
...
 
 You can then run a Docker container for the API in one of three ways:
-#### Option 1: Use the `docker-compose.yaml` file
+#### Option 1 (RECOMMENDED): Use the `docker-compose.yaml` file
 
 First, [install docker compose](https://docs.docker.com/compose/install/).
 
 If needed, update your `.env` file with optional environment variables for the docker compose configuration.
 
+**TIP:** Double check that the environment variables are resolved with your expected values using the command `docker compose config`.
```
surchs, line 98: nice!

README.md
```diff
@@ -56,60 +56,71 @@ NB_GRAPH_USERNAME=someuser
...
 docker compose up -d
 ```
 
 #### Option 2: Use the latest image from Docker Hub
 ```bash
-export $(cat .env | xargs)  # export your environment variables 
+export $(cat .env | xargs)  # OPTIONAL: export your environment variables
 docker pull neurobagel/api
-docker run -d --name api -p 8000:8000 --env NB_GRAPH_USERNAME --env NB_GRAPH_PASSWORD neurobagel/api
+docker run -d --name=api -p ${NB_API_PORT_HOST:-8000}:${NB_API_PORT:-8000} --env-file=.env neurobagel/api
```
surchs, line 109: I get what you have in mind here - but I don't think the README section for "minimal command to run this" is the right place for this relatively complex syntax. As far as I understand, the motivation for `NB_API_PORT_HOST` and `NB_API_PORT` is when I use `docker compose`, yes? in that case I'd just stick with a literal port value - we can still mention that this can be changed.

app/api/crud.py
```diff
@@ -85,6 +85,7 @@ async def get(
                 os.environ.get(util.GRAPH_USERNAME.name),
                 os.environ.get(util.GRAPH_PASSWORD.name),
             ),
+            timeout=10.0,
```
surchs, line 88: 1. I think this means you can also close https://github.com/neurobagel/api/issues/140 with this
2. I also think that 10 seconds is not enough given the query times we have already seen - until we fix the underlying issue of unoptimized and slow queries, let's not set this value too low. How about a) a TODO comment to revisit the value when performance improves, b) 30 seconds?

surchs (changes requested):
Thanks for addressing this @alyssadai. I think this is good to go, but I have not yet approved it because I have some questions about the Dockerfile and the API port setting changes you made. It's possible (likely?) that I just don't understand all you are doing here but I think it'd be good to make sure first. 

Take a look

alyssadai:
Hey @surchs, thanks again for your detailed review of this.

In the README, I updated the `docker run` commands/corresponding instructions in the option 2 and option 3 subsections of the Docker installation section based on your comments. I also added a note to the Dockerfile about how the environment variable for the API port behaves.

Finally, I've added a commented template .env file and instructions to use it in the README.

PR #151, opened 2023-08-29
Add environment variable for allowed origins for CORS

### LINKED ISSUE #150: Make allowed origins for CORS configurable
- [x] create environment variable for the list of trusted origins for which CORS is allowed
- [x] update docker-compose.yml to define new env var in `environment`
- [x] update README to include examples for configuring this list
  - e.g. for an internal deployment, the environment variable should include the localhost port the query tool is running on
- [x] update `.template-env`
- [x] add warning when no allowed origins are set

Test that new environment variable is parsed correctly:

- [x] empty list -> allowed origins should be empty list
- [x] undefined environment variable -> allowed origins should be empty list (no error)
- [x] more than one origin -> allowed origins should be a list containing all

alyssadai:
Closes #150

- New environment variable `NB_API_ALLOWED_ORIGINS` replaces hard-coded wildcard allowing all origins for CORS
  - default value is an empty string
- Add utility function to parse origins into a list
- Raise warning if no allowed origins are specified
- Tests added that specified origins are parsed correctly + warning raised appropriately

## Checklist

- [ ] PR has an interpretable title with a prefix (`[ENH]`, `[FIX]`, `[REF]`, `[TST]`, `[CI]`, `[MNT]`, `[INF]`, `[MODEL]`, `[DOC]`) _(see https://neurobagel.org/contributing/pull_requests for more info)_
- [ ] PR links to GitHub issue with mention `Closes #XXXX`
- [ ] Tests pass
- [ ] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:

### DIFF
--- a/.template-env
+++ b/.template-env
@@ -4,6 +4,7 @@ NB_GRAPH_PASSWORD=DBPASSWORD  # REPLACE DBPASSWORD WITH YOUR GRAPH DATABASE PASS
 NB_GRAPH_DB=test_data/query
 NB_RETURN_AGG=true
 NB_API_TAG=latest
+NB_API_ALLOWED_ORIGINS="https://localhost:3000 http://localhost:3000"
 NB_GRAPH_IMG=stardog/stardog:8.2.2-java11-preview
 ## ADDITIONAL CONFIGURABLE PARAMETERS: Uncomment and modify values of the below variables as needed to use non-default values.
 # NB_API_PORT_HOST=8000
--- a/README.md
+++ b/README.md
@@ -24,6 +24,7 @@ The Neurobagel API is a REST API, developed in [Python](https://www.python.org/)
 - [Quickstart](#quickstart)
 - [Local installation](#local-installation)
     - [Environment variables](#set-the-environment-variables)
+    - [Using a graphical query tool](#using-a-graphical-query-tool-to-send-api-requests)
     - [Docker](#docker)
     - [Python](#python)
 - [Testing](#testing)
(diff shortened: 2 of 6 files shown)


.template-env
```diff
@@ -4,6 +4,7 @@ NB_GRAPH_PASSWORD=DBPASSWORD  # REPLACE DBPASSWORD WITH YOUR GRAPH DATABASE PASS
 NB_GRAPH_DB=test_data/query
 NB_RETURN_AGG=true
 NB_API_TAG=latest
+NB_API_ALLOWED_ORIGINS="https://localhost:3000 http://localhost:3000"
```
surchs, line 7: If I understand correctly, you want to show an example of two origins here? That makes sense to me. This is a minor comment, but I think it'd be better to use something like 
```suggestion
NB_API_ALLOWED_ORIGINS="https://localhost:3000 https://myOtherOrigin"
```
instead for the second option to highlight that we aren't advising people to configure their API for both HTTP and HTTPS. If they configure HTTPS, they likely also have NGINX setup, so they may not even need to configure CORS at all. For the folks who just quickly follow the defaults, best to keep it simple. Maybe a comment helps

README.md
```diff
@@ -84,6 +86,29 @@ also ensure that any variables defined in your `.env` are not already set in you
...
+This variable defaults to an empty string (`""`) when unset, meaning that your deployed API will only accessible via direct `curl` requests to the URL where the API is hosted (see [this section](#send-a-test-query-to-the-api) for an example `curl` request).
+
+Note: The `.template-env` file in this repo assumes you want to allow API requests from a query tool hosted at a specific port on `localhost`.
+
+Other examples:
+```bash
+# ---- .env file ----
+
+# do not allow requests from any frontend origins
+NB_API_ALLOWED_ORIGINS=""  # or, exclude variable from .env
```
surchs, line 100:
```suggestion
NB_API_ALLOWED_ORIGINS=""  # this is the default value that will also set if you exclude the variable from the .env file
```

tests/test_query.py
```diff
@@ -114,6 +117,72 @@ def mock_httpx_post(**kwargs):
...
+    monkeypatch.setenv(util.GRAPH_PASSWORD.name, "DBPASSWORD")
+
+    with pytest.warns(
+        UserWarning,
+        match=f"API was launched without providing any values for the {util.ALLOWED_ORIGINS.name} environment variable",
+    ):
+        with test_app:
+            pass
+
+    assert util.parse_origins_as_list(
```
surchs, line 134: Why is this part of the test? Aren't you implicitly testing the monkeypatch function here? For the `parse_origins_as_list` portion the next test is a better fit.

tests/test_query.py
```diff
@@ -114,6 +117,72 @@ def mock_httpx_post(**kwargs):
...
+    Test that when the environment variable for allowed origins has been explicitly set, the app correctly parses it into a list
+    and raises a warning if the value is an empty string.
+    """
+    monkeypatch.setenv(util.ALLOWED_ORIGINS.name, allowed_origins)
+    # set random username and password to avoid RuntimeError from other startup check
+    monkeypatch.setenv(util.GRAPH_USERNAME.name, "DBUSER")
+    monkeypatch.setenv(util.GRAPH_PASSWORD.name, "DBPASSWORD")
+
+    with expectation:
+        with test_app:
```
surchs, line 175: I know this isn't a new pattern but I forgot why we do that. Since I also couldn't find it in our internal wiki, would you mind reminding me why the `with test_app:` bit is needed to catch exceptions and then add it to the wiki, maybe here: https://github.com/neurobagel/documentation/wiki/API ?

surchs (changes requested):
Thanks for this PR @alyssadai. Everything functional is looking good! I still requested changes to make sure we communicate our expectations for CORS clearly to the user - I left some minor comments on where I think we should clarify things.

Take a look and let me know what you think. Then I think this can go

.template-env
alyssadai: Okay, that's a fair point regarding http/https.

I wanted to provide a sensible template value that the user would not have to edit if following our recommended setup with a locally deployed query tool. In this case since I wasn't sure if they would have HTTPS configured I listed both to cover all the bases.

Instead of `https://myOtherOrigin` as the second origin (which a user can't use as-is), how about `https://127.0.0.1:3000` to cover both ways of accessing the query tool on localhost? Alternatively, we could have `https://query.neurobagel.org` as the second origin, but I'm not sure if users would want to make a local API accessible by our tool by default.

surchs: I see. OK, then how about this: keep your original version, just a comment above to say something like:
`# template for adding multiple origins, please edit this line for your own setup before deploying`?

app/main.py
```diff
@@ -33,6 +35,17 @@ async def auth_check():
         )
 
 
+@app.on_event("startup")
```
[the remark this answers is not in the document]
alyssadai: Tracked in https://github.com/neurobagel/api/issues/152

tests/test_query.py
alyssadai: The purpose of this test is to check, when the `NB_API_ALLOWED_ORIGINS` environment variable is undefined, that the API: 
- a) emits a warning (this warning is also emitted when `NB_API_ALLOWED_ORIGINS` is explicitly set to `""` by the user)
- b) internally assigns a default value of `""` for the variable, meaning the CORS middleware ends up with `[""]` as the list of allowed origins

The assertion above is to explicitly check b), that is, that the final list of origins isn't something weird just because the environment variable wasn't set.

tests/test_query.py
alyssadai: This pattern is actually documented in FastAPI's own docs, https://fastapi.tiangolo.com/advanced/testing-events/ :)

surchs: heyho, that's great. Thanks, I missed that

surchs (approved):
looks good @alyssadai, :cook:

PR #164, opened 2023-09-06
[FIX] Remove yaml list key for API_QUERY_URL

### LINKED ISSUE #163: docker compose env var for shell substitution is incorrect
https://github.com/neurobagel/api/blob/28f1b0e3cf560d866baebf124c7a880cae515bab/docker-compose.yml#L27

uses a `-` dash, but: https://docs.docker.com/compose/environment-variables/set-environment-variables/#substitute-from-the-shell seems to suggest plain key-value syntax. 

in the current form, `docker compose up` raises the following error:

```
validating /home/surchs/Repositories/MNI/bagelapi/docker-compose.yml: services.query.environment.0 must be a string
```

We should:

- confirm that the key-value syntax is required and the array syntax is not allowed
- remove the dash
- make a note of that for other `docker compose` files

surchs:
Closes #163

- remove the dash

Possibly still to be done:
- confirm this is actually a yaml-list (i.e. `-` dash) issue
- can we test this so it doesn't happen again?

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -24,4 +24,4 @@ services:
     ports:
       - "${NB_QUERY_PORT_HOST:-3000}:3000"
     environment:
-      - API_QUERY_URL: "${API_QUERY_URL:-http://localhost:${NB_API_PORT_HOST:-8000}/}"
+      API_QUERY_URL: "${API_QUERY_URL:-http://localhost:${NB_API_PORT_HOST:-8000}/}"


alyssadai (approved):
Thanks for the quick fix @surchs 💪

Re: 

> can we test this so it doesn't happen again?

The easiest solution would probably be to add to our pre-commit config a hook to validate the docker compose config file itself, i.e.
- `docker-compose-check` - https://github.com/IamTheFij/docker-pre-commit (also listed under https://pre-commit.com/hooks.html)

surchs:
> Thanks for the quick fix @surchs 💪
> 
> 
> 
> Re: 
> 
> 
> 
> > can we test this so it doesn't happen again?
> 
> 
> 
> The easiest solution would probably be to add to our pre-commit config a hook to validate the docker compose config file itself, i.e.
> 
> - `docker-compose-check` - https://github.com/IamTheFij/docker-pre-commit (also listed under https://pre-commit.com/hooks.html)
> 
> 

Yeah I think that's a really good idea, could you make an issue and then we address that maybe for the dsst milestone?

PR #167, opened 2023-09-07
[FIX] Properly expose `NB_GRAPH_PORT_HOST` to API container

### LINKED ISSUE #166: `NB_GRAPH_PORT_HOST` not exposed to API environment in Docker Compose
The API currently uses the environment variable `NB_GRAPH_PORT` to define the URL for the graph db to be queried,
https://github.com/neurobagel/api/blob/126e5acfdeebae3d9632858992e5651ff76ae990/app/api/utility.py#L26-L34

but this variable is not defined in the environment for the API container in the docker compose config (only for the graph):

https://github.com/neurobagel/api/blob/126e5acfdeebae3d9632858992e5651ff76ae990/docker-compose.yml#L8-L15

this means that even when `NB_GRAPH_PORT` has been set by the user, the API defaults to a value of `5820` which results in strange behaviour when sending requests using a locally deployed query tool (for some reason, there is both a CORS error and an internal server error???)

To fix, we need to:
- [ ] change API to use `NB_GRAPH_PORT_HOST` instead of `NB_GRAPH_PORT` since the former is what's actually exposed to the network
- [ ] expose `NB_GRAPH_PORT_HOST` to the API in `docker-compose.yml`.

alyssadai:
Closes #166

- `NB_GRAPH_PORT` variable previously erroneously used by API also switched to `NB_GRAPH_PORT_HOST` (since this is what's exposed)

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:


rmanaem (approved):
:cook:

Issue #165, opened 2023-09-06
Add pre-commit hook to check docker-compose syntax


alyssadai:
To hopefully complain if we have extra dashes, missing quotes, etc.

`docker-compose-check` - https://github.com/IamTheFij/docker-pre-commit (also listed under https://pre-commit.com/hooks.html)

PR #177, opened 2023-09-18
[FIX] Removed the nested variable for `API_QUERY_URL` env var

### LINKED ISSUE #170: [BUG] nested variable substitution fails

- [ ] I have searched the existing issues

### Expected Behavior

For the default docker-compose.yml file to work without any modifications

### Current Behavior

Nested variable substitution doesn't work for my version of Docker Compose version v2.2.3

So the original config:

    environment:
      API_QUERY_URL: "${API_QUERY_URL:-http://localhost:${NB_API_PORT_HOST:-8000}/}"

Needs to be:

    environment:
      API_QUERY_URL: "${API_QUERY_URL:-http://localhost:8000/}"

### Error message

invalid interpolation format for services.query.environment.API_QUERY_URL: "[http://localhost:${NB_API_PORT_HOST:-8000](http://localhost:%24%7Bnb_api_port_host:-8000/)". You may need to escape any $ with another $.

### Environment

- OS:
- Python/Node version:

### How to reproduce

_No response_

### Anything else?

_No response_

rmanaem:
Closes #170 

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -25,4 +25,4 @@ services:
     ports:
       - "${NB_QUERY_PORT_HOST:-3000}:3000"
     environment:
-      API_QUERY_URL: "${API_QUERY_URL:-http://localhost:${NB_API_PORT_HOST:-8000}/}"
+      API_QUERY_URL: "${API_QUERY_URL:-http://localhost:8000/}"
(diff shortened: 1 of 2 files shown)


alyssadai (approved):
Nice! Thanks for addressing this so quickly @rmanaem.

One request and a 🍒:
- In the environment variable TSV in this repo, can you update the Description for `API_QUERY_URL` to mention that the port number in the URL must correspond to `NB_API_PORT_HOST` (if the user is not using the default)?
- 🍒: this is optional, but I realized that issue https://github.com/neurobagel/api/issues/165 could fit well into this PR since the pre-commit hook I mention there would error out with the old syntax we are fixing (just tested on the server). So this would essentially add a test that would fail with the old behaviour 👀. What do you think about tacking that onto this PR?

rmanaem:
>🍒: this is optional, but I realized that issue https://github.com/neurobagel/api/issues/165 could fit well into this PR since the pre-commit hook I mention there would error out with the old syntax we are fixing (just tested on the server). So this would essentially add a test that would fail with the old behaviour 👀. What do you think about tacking that onto this PR?

I think It's better to address this in another PR.

PR #183, opened 2023-09-22
[ENH] Implemented endpoint for returning available termURLs

### LINKED ISSUE #37: Endpoint that will return the list of available termURLs for a given query parameter
There should be an API endpoint or resource where I can request a list of all of the termURLs that are currently available for a given variable in the graph, so that I know what makes sense to query.

e.g. I want to get all of the unique `diagnosis` terms that exist in the current graph returned as an array.

~There might also be a way to just get this for all of the categorical variables: e.g. a single endpoint that returns an object with the existing terms for all variables:~ (EDIT: we will not implement this as one endpoint - instead we will have a separate endpoint that returns all the variables with controlled term instances. See also #176)
```json
{
"diagnosis": ["snomed:1", "snomed:2", ...],
"imagingModality": ["T1", "T2", ...],
...
}
```

### Steps to implement
- [x] Create a `GET` path operation with the controlled term of the Neurobagel class (e.g., `nb:Assessment`?) as a [path parameter](https://fastapi.tiangolo.com/tutorial/path-params/)
  - `api.neurobagel.org/attributes/{attribute_URI}`
  - e.g. route for getting all the TermURLs for Diagnosis would be `api.neurobagel.org/attributes/nb:Diagnosis`
    - note: having colons in the path param _should_ be fine, see https://github.com/encode/starlette/discussions/1673
- [x] corresponding crud function will post a query to the graph endpoint to return all instances of that class
- [x] Response should be in the format of:
`{attribute_URI: [termURL1, termURL2, termURL3…]}`

References for inspo:
- https://www.cognitiveatlas.org/api
- GraphDB [retrieving namespaces](https://rdf4j.org/documentation/reference/rest-api/#tag/Repositories/paths/~1repositories/get) vs [getting a namespace by prefix](https://rdf4j.org/documentation/reference/rest-api/#tag/Namespaces/paths/~1repositories~1%7BrepositoryID%7D~1namespaces~1%7BnamespacesPrefix%7D/get)

rmanaem:
Closes #37 

- Implemented `create_terms_query` function in `utility.py`
- Implemented `get_terms` function in `crud.py`
- Implemented `get_terms` path operation in `query.py`
- Moved `pytest.fixtures` from `test_query.py` to `conftest.py`
- Implemented tests for valid and invalid `attribute_URI` and the `get_terms` path operation

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [x] Tests have been added

For bug fixes:


app/api/crud.py
```diff
@@ -151,3 +151,44 @@ async def get(
...
+        raise HTTPException(
+            status_code=response.status_code,
+            detail=f"{response.reason_phrase}: {response.text}",
+        )
+
+    results = response.json()
+
+    results_dict = {
+        attribute_URI: [
+            i["termURL"]["value"] for i in results["results"]["bindings"]
```
surchs, line 190:
```suggestion
            result["termURL"]["value"] for result in results["results"]["bindings"]
```
:cherries: but clearer for sanity

app/api/routers/query.py
```diff
@@ -25,3 +26,11 @@ async def get_query(query: QueryModel = Depends(QueryModel)):
     )
 
     return response
+
+
+@router.get("/attributes/{attribute_URI}")
```
surchs, line 31:
```suggestion
@router.get("/attributes/{dataElementURI}")
```
I don't think we're fully settled yet on what we want to call these things, but attribute is probably not the right word. Either "variable" or "common data element" or "data element" makes more sense

app/api/utility.py
```diff
@@ -205,3 +205,28 @@ def create_query(
...
+
+
+def create_terms_query(attribute_URI: str) -> str:
+    """
+    Creates a SPARQL query using a simple query template to retrieve term URLS for a given attribute.
+
+    Parameters
+    ----------
+    attribute_URI : str
+        The URI of the attribute for which to retrieve term URLS.
```
surchs, line 217:
```suggestion
        The URI of the data element for which to retrieve the URIs of all connected term.
```

app/api/utility.py
```diff
@@ -205,3 +205,28 @@ def create_query(
         """
 
     return "\n".join([DEFAULT_CONTEXT, query_string])
+
+
+def create_terms_query(attribute_URI: str) -> str:
+    """
+    Creates a SPARQL query using a simple query template to retrieve term URLS for a given attribute.
+
+    Parameters
```
surchs, line 214: I think this docstring could benefit from an example: https://numpydoc.readthedocs.io/en/latest/format.html#examples so we can immediately understand what we mean by `dataElementURI` or similar

tests/conftest.py
```diff
@@ -8,3 +8,116 @@
 def test_app():
     client = TestClient(app)
     yield client
+
+
+@pytest.fixture()
+def test_data():
```
surchs, line 14: I don't think these fixtures will be used anywhere but the `test_query` tests for now, no? Is the only reason you're refactoring them here to make the actual test module shorter? I find it a bit less readable to hide them if the only thing we gain from this is a shorter test file

tests/test_query.py
```diff
@@ -473,3 +394,33 @@ async def mock_get(
...
+@pytest.mark.parametrize(
+    "invalid_attribute_URI",
+    ["apple", "some_thing:cool"],
+)
+def test_get_terms_invalid_attribute_URI(
+    test_app, mock_invalid_get_terms, invalid_attribute_URI, monkeypatch
+):
+    """Given a valid attribute URI, returns a 422 status code."""
+
+    monkeypatch.setattr(crud, "get_terms", mock_invalid_get_terms)
```
surchs, line 424: why do I need to mock this at all? If I understand correctly, the error here should come from the validation of the term (i.e. doesn't fit Regex). So I would expect that the API sends me this error back before a get request is ever made.

Could you check this again / see if the test passes without mocking get at all?

surchs (commented):
Hey @rmanaem, thanks for the PR. I left some questions in the code below, take a look and let me know what you think. I think for clarity's sake, we should rename `attribute_URI` to `nb:DataElementURI` or something similar. 

`nb:Diagnosis` is not an attribute, but a `variable` or `commonDataElement`

tests/test_query.py
rmanaem: Yea there is no need to mock that there.

surchs: I would then also update the docstring to explain that it is supposed to trigger the validation, which in turns should return the 422 code. i.e. explaining the "why" here is what will help in the future

tests/conftest.py
rmanaem: Yea kinda. We have quite a few tests and couple fixtures in one module and it seemed cluttered, I thought it's better to move the fixtures to `conftest.py` as I was going to add a few more and make it even more cluttered.
While the pytest docs suggests moving fixtures to `conftest.py` for an entire directory to have access to (which we don't need atm), I think it simplifies the `test_query` module and offers clear test structure.

app/api/utility.py
```diff
@@ -205,3 +205,31 @@ def create_query(
...
+    data_element_URI : str
+        The URI of the data element for which to retrieve the URIs of all connected term.
+
+    Returns
+    -------
+    str
+        The SPARQL query.
+    Examples
+    --------
+    get_terms_query("nb:Assessment")
```
surchs, line 225: neat!

surchs (approved):
Thanks, looks great! :cook:

PR #191, opened 2023-09-27
[FIX] Properly expose `NB_GRAPH_PORT` to API container 

### LINKED ISSUE #190: [BUG] Have to use `NB_GRAPH_PORT` instead of `NB_GRAPH_PORT_HOST`

- [x] I have searched the existing issues

### Expected Behavior

At the moment, we're using `NB_GRAPH_PORT_HOST` to construct the Query URL. However the API is in a docker network with the graph and will therefore have to send requests to the graph on the graph's internal port `NB_GRAPH_PORT`. 

https://github.com/neurobagel/api/blob/2a6fc31bc4bc11ee767860d294fc076f7d9c2e73/app/api/utility.py#L26-L36

We have to change these lines and the corresponding docker-compose file so that we expose and use `NB_GRAPH_PORT` instead.

### Current Behavior

_No response_

### Error message

_No response_

### Environment

_No response_

### How to reproduce

_No response_

### Anything else?

We should also document in our internal docs that within a docker network services talk to each other on their internal port number, and not the one that is exposed to the host.

rmanaem:
Closes #190 

- Changed `NB_GRAPH_HOST_PORT` env var to `NB_GRAPH_PORT` in `utility.py` and `docker-compose.yml`
- See [internal port vs host port](https://github.com/neurobagel/documentation/wiki/Docker#internal-port-vs-host-port) for internal wiki entry

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:


alyssadai (approved):
🧑‍🍳

PR #196, opened 2023-10-04
[ENH] Update `add_data_to_graph.sh` to also upload TTL files

### LINKED ISSUE #185: Expand functionality of data uploader script to upload both Turtle and JSONLD files in directory
We are starting to create + think about vocabulary graph files which are much more easily expressed in a `.ttl` file than `.jsonld` (e.g., https://github.com/neurobagel/api/issues/182), but nonetheless need to have a way to be easily added to a graph.

Currently, [add_data_to_graph.sh](https://github.com/neurobagel/api/blob/main/add_data_to_graph.sh) only uploads `.jsonld` files in a directory. To make it easy for a deploying user to upload _all data_ that needs to be in the graph in one go, we could add another loop to the script that performs the same `POST` request on any `.ttl` files in the specified directory.

https://github.com/neurobagel/api/blob/b65a2428be168150d9139fb7d2de2b115f146763/add_data_to_graph.sh#L146-L150 

This script would then work on both directories with just one kind of file (`.ttl` or `.jsonld`) or those with a mix of both - allowing us flexibility if we want to recommend storing all vocab-related files separately but still be able to merge them into the graph using the same script.

We should then also update the deployment docs to mention how to add vocabulary files to the graph.

alyssadai:
Closes #185 

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:

### DIFF
--- a/add_data_to_graph.sh
+++ b/add_data_to_graph.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 #
 # ARG_HELP([Upload JSONLD and Turtle data to a Neurobagel graph])
-# ARG_POSITIONAL_SINGLE([dir],[Path to directory containing .jsonld and/or .ttl files to be uploaded])
+# ARG_POSITIONAL_SINGLE([dir],[Path to directory containing .jsonld and/or .ttl files. ALL .jsonld and .ttl files in this directory will be uploaded.])
 # ARG_POSITIONAL_SINGLE([graph-url],[Host and port at which to access the graph database to add data to (e.g., localhost:7200)])
 # ARG_POSITIONAL_SINGLE([graph-db],[Name of graph database to add data to])
 # ARG_POSITIONAL_SINGLE([user],[Username for graph database access])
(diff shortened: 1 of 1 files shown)


add_data_to_graph.sh
```diff
@@ -39,9 +39,9 @@ _arg_clear_data="off"
 
 print_help()
 {
-	printf '%s\n' "Upload JSONLD data to a Neurobagel graph"
+	printf '%s\n' "Upload JSONLD and Turtle data to a Neurobagel graph"
 	printf 'Usage: %s [-h|--help] [--(no-)clear-data] <dir> <graph-url> <graph-db> <user> <password>\n' "$0"
-	printf '\t%s\n' "<dir>: Path to directory containing .jsonld files to be uploaded"
+	printf '\t%s\n' "<dir>: Path to directory containing .jsonld and/or .ttl files to be uploaded"
```
surchs, line 44: Not a new thing, but: maybe a note that ALL files will be uploaded that are at this path.

surchs (approved):
1 comment, and then :cook:

PR #205, opened 2023-10-19
[ENH] Make data upload script backend responsive

### LINKED ISSUE #204: Add flag to `add_data_to_graph.sh` for graph backend and increase verbosity
- [x] Add a boolean argument to the script where one condition is Stardog and the other is GraphDB, and internally adjust the graph endpoints for the `curl` requests as necessary via variables
- [x] In the script, capture response codes and output to user a list of ones that don't have code 2xx during the upload

alyssadai:
Closes #204

- Add bool flag for whether to use GraphDB API endpoints (if off, uses Stardog endpoints)
- For each request to the graph (clearing data or uploading data), explicitly capture response codes and increase verbosity for easier debugging
  - If database clearing step fails (when selected), exit script instead of continuing to upload the data files
  - At end of process, output to user a list of any db filenames that didn't succeed in uploading (status code != 2xx)
- Capitalize script output messages for easier visual differentiation from HTTP error messages

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:


rmanaem (approved):
The changes look good.
However, I do have a question and a suggestion.
I'm curious to know why you capitalized the messages? Is it a convention?
:cherries: : I think it's a good idea to rename either `graph-db` or `use-graphdb-syntax` in code for the sake of clarity e.g., `use-GraphDB-syntax`.

rmanaem (approved):
The changes look good. 
However, I do have a question and a suggestion.
I'm curious to know why you capitalized the messages? Is it a convention?
:cherries: : I think it's a good idea to rename either `graph-db` or `use-graphdb-syntax` in code for the sake of clarity/readability e.g., `use-GraphDB-syntax`.
Other than that :cook:

alyssadai:
> I'm curious to know why you capitalized the messages? Is it a convention?

Mostly so that it is easy for the user to visually distinguish the progress messages from the script itself (e.g. `CLEARING EXISTING DATA ...`) and the HTTP responses returned from the actual graph API which are also printed to the console (since all the curl commands for uploading data are called in a way as to output the response headers for debugging). Otherwise, the output from the script may be a bit tricky to parse.

> 🍒 : I think it's a good idea to rename either graph-db or use-graphdb-syntax in code for the sake of clarity e.g., use-GraphDB-syntax.

Good point! I've renamed the `graph-db` parameter to `graph-database` 🙂

PR #214, opened 2023-11-06
[MNT] Removed `session_file_path` from aggregate results

### LINKED ISSUE #210: Remove session path values from aggregate results
Since the session paths we're currently returning contain BIDS subject identifiers, a valid concern is that even these paths are exposing too much identifying info for a supposedly 'secure' aggregate response.

To be safe, we will turn this off for the aggregate mode of the API so that the graph no longer returns the session paths.

See also #208

Looking ahead:
- When we eventually make the switch to having a two-part download process (for the dataset-level vs. participant-level TSV), we should consider not returning the participant-level TSV at all when the API is deployed in aggregate mode

rmanaem:
Closes #210 

- Updated `create_query` function to not query `session_file_path`

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:

### DIFF
--- a/app/api/crud.py
+++ b/app/api/crud.py
@@ -138,10 +138,7 @@ async def get(
             by=dataset_cols
         ):
             if util.RETURN_AGG.val:
-                subject_data = list(
-                    {"session_file_path": file_path}
-                    for file_path in group["session_file_path"].dropna()
-                )
+                subject_data = "protected"
             else:
                 subject_data = (
                     group.drop(dataset_cols, axis=1)
--- a/app/api/models.py
+++ b/app/api/models.py
@@ -1,7 +1,7 @@
 """Data models."""
 
 from enum import Enum
-from typing import Optional
+from typing import Optional, Union
 
 from fastapi import Query
 from fastapi.exceptions import HTTPException
@@ -60,7 +60,7 @@ class CohortQueryResponse(BaseModel):
     dataset_portal_uri: Optional[str]
     records_protected: bool
     num_matching_subjects: int
-    subject_data: list[dict]
+    subject_data: Union[list[dict], str]
     image_modals: list
 
 
--- a/app/api/utility.py
+++ b/app/api/utility.py
@@ -213,9 +213,9 @@ def create_query(
     # wrap query in an aggregating statement so data returned from graph include only attributes needed for dataset-level aggregate metadata.
     if return_agg:
         query_string = f"""
-            SELECT ?dataset_uuid ?dataset_name ?dataset_portal_uri ?sub_id ?session_file_path ?image_modal WHERE {{\n
+            SELECT ?dataset_uuid ?dataset_name ?dataset_portal_uri ?sub_id ?image_modal WHERE {{\n
             {query_string}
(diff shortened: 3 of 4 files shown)


alyssadai (commented):
Thanks for the PR @rmanaem. I have a few implementation questions:
- A quick test of a local aggregate-mode API with your changes shows that the `subject_data` field still appears in the response, just as an empty list `[]`. Should we remove this key entirely from the aggregate response to avoid API user confusion? Or maybe populate it with an informative value like `"protected"`? 
- I noticed that this logic is still kicking around - not sure if we need it anymore
https://github.com/neurobagel/api/blob/47d3b7e720d29a9fd9b02f25343307e4d6bc6ed2/app/api/crud.py#L140-L144
- I see you reverted your initial change to the `test_data` aggregate response fixture, but I am in favour of updating it to reflect what the actual aggregate response looks like - wdyt?

Happy to have a chat about these decisions if that'd be helpful.

rmanaem:
> A quick test of a local aggregate-mode API with your changes shows that the subject_data field still appears in the response, just as an empty list []. Should we remove this key entirely from the aggregate response to avoid API user confusion? Or maybe populate it with an informative value like "protected"?

I like the `protected` idea but to just to be sure is this `subject_data: ["protected"]` what you had in mind?

> I noticed that this logic is still kicking around - not sure if we need it anymore

Great catch, I'll get rid of that.

> I see you reverted your initial change to the test_data aggregate response fixture, but I am in favour of updating it to reflect what the actual aggregate response looks like - wdyt?

I don't have a preference but If we were to do that I'd say both subjects in the `test_data` should have their `subject_data` updated instead of just one like my commit did. wdyt?

alyssadai:
> I like the `protected` idea but to just to be sure is this `subject_data: ["protected"]` what you had in mind?

I _think_ so, although you probably have a better sense of what would work best with the query tool. Since the API currently always expects a list of dicts for `subject_data`,
https://github.com/neurobagel/api/blob/47d3b7e720d29a9fd9b02f25343307e4d6bc6ed2/app/api/models.py#L63

I'm even wondering if we should just return `subject_data: "protected"` and make the model accept either a dict or a string value for this key. Alternatively, as said above, we could make this key entirely optional in the model.

> I don't have a preference but If we were to do that I'd say both subjects in the test_data should have their subject_data updated instead of just one like my commit did. wdyt?

Yes, since the fixture represents a single response I think all the subjects should be updated!

rmanaem:
> I think so, although you probably have a better sense of what would work best with the query tool. Since the API currently always expects a list of dicts for subject_data,

Yea that's why I prefer the not removing the `subject_data` field. The query tool checks for that so I'd rather have it in the response even if it's empty.

> I'm even wondering if we should just return subject_data: "protected" and make the model accept either a dict or a string value for this key. Alternatively, as said above, we could make this key entirely optional in the model.

I had completely forgotten about that. I like the idea of extending the model to accept string values as well.

alyssadai (approved):
I think this looks good! 🧑‍🍳

PR #218, opened 2023-11-09
[ENH] Return human-readable labels (if available) as well as term IDs from `/attributes/{data_element_URI}`

### LINKED ISSUE #217: Update `/attributes/{data_element_URI}` to return both terms AND labels
- Should be able to use the term lookup file created by API on startup for each vocab (for now, just the Cognitive Atlas)

Response from the endpoint will change from this:
```json
{
  "nb:Diagnosis": [
     "snomed:trm12345",
     "snomed:trm23456",
     "snomed:trm34567",
  ] 
}
```

to this:
```json
{
  "nb:Diagnosis": [
     {"TermURL": "snomed:trm12345", "Label": "my term 1"},
     {"TermURL": "snomed:trm23456", "Label": "my term 2"},
     {"TermURL": "snomed:trm34567", "Label": "my term 3"},
  ] 
}
```

alyssadai:
- Closes #217 

Endpoint now returns a dictionary in this format, representing terms of a given class that are used in the graph:
```json
{
  "nb:Assessment": [
     {"TermURL": "cogatlas:trm12345", "Label": "my term 1"},
     {"TermURL": "cogatlas:trm23456", "Label": "my term 2"},
     {"TermURL": "cogatlas:trm34567", "Label": "my term 3"},
  ]
}
```

New logic has been added such that:
  - if a found term for a given data element has an **unrecognized namespace** in it, it will not be processed + returned to the user by the endpoint (but API will emit a warning in the console)
  - if the data element passed to this endpoint **does not have a vocabulary lookup** available (i.e., nowhere to check term-label mappings), just return the IDs of the term instances themselves
  - if a found term for a given data element **has a vocabulary lookup available, but the term itself is not found in the lookup file**, just return the term ID

and tests!!!

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:


app/api/crud.py
```diff
@@ -177,28 +178,62 @@ async def get(
...
+        # First, check whether the found instance of the data element contains a recognized namespace
+        if util.is_term_namespace_in_context(term_url):
+            # Then, check if we have a vocab lookup available to get the human-readable label for the term
+            if term_labels_path is not None:
+                # If so, attempt to get the label for the term
+                term_label = vocab_term_lookup.get(
+                    util.strip_namespace_from_term_uri(term_url), None
+                )
+            else:
+                term_label = None
```
surchs, line 220: could we replace that by having an `else` statement above for when `term_labels_path is not None` and then just setting `vocab_term_lookup = {}`?

app/api/utility.py
```diff
@@ -251,7 +251,56 @@ def create_terms_query(data_element_URI: str) -> str:
...
+    Returns
+    -------
+    str
+        The unique term ID.
+    """
+    if has_prefix:
+        for prefix in CONTEXT:
+            if prefix in term:
+                return term.replace(prefix, "")
+    elif not has_prefix:
```
surchs, line 294: :cherries: 
```suggestion
    else:
```

app/api/utility.py
```diff
@@ -251,7 +251,56 @@ def create_terms_query(data_element_URI: str) -> str:
...
+        A controlled term URI.
+    has_prefix : bool, optional
+        Whether the term URI includes a namespace prefix (as opposed to the full namespace URI), by default False.
+
+    Returns
+    -------
+    str
+        The unique term ID.
+    """
+    if has_prefix:
```
surchs, line 290: if we already know that this has a prefix, would there ever be a case where we have more than one `:` (e.g. `cogatlas:MyHappyValueIs:42`)? Because if not, why don't we just split the prefix off and then search for it directly?

surchs (approved):
nice PR @alyssadai :tada: !

I left some smaller comments, take a look. And then :cook:

PR #228, opened 2023-11-23
[ENH] Add support for endpoint that will return `nb:Diagnosis` options from SNOMED

### LINKED ISSUE #211: Update vocabulary path to parse and return `nb:Diagnosis` options
Add workflow for the path `/attributes/nb:Diagnosis/vocab` that:
- [x] Reads in a file of SNOMED diagnosis term-label mappings (downloaded from Athena) on startup + creates a lookup file
  - a copy should probably be stored in the repo for now, under https://github.com/neurobagel/api/tree/main/vocab/backup_external
  - The diagnosis term-label mappings can be found here https://github.com/neurobagel/annotation_tool/blob/main/static/diagnosisTerms.json
- [x] Processes the mappings into a `VocabLabelsResponse` object
- [x] append the missing terms (see below) to the SNOMED-CT .json dump
  - already done in .json used by annotation tool
- [x] update the code snippet in https://github.com/neurobagel/planning/issues/52 to also include the appended terms that are missing, so it is re-executable
- ~Allows only password-authenticated users to access that endpoint~

Other TODO:
- [x] wrap requests to external vocab APIs in a `try-except` for httpx network errors (currently, if the network connection cannot be established if, e.g. the internet is down, an exception is raised and the app cannot start)
  - [x] if catch is hit, raise informative warning and use local copy of vocab file

alyssadai:
- Closes #211 

- Added JSON dump of SNOMED CT disorder terms
- On startup, app will now read in the file of SNOMED diagnosis terms and store a lookup file where entries are in the format `<id>:<name>`
- Implemented logic for `/attributes/nb:Diagnosis/vocab` which loads + processes the lookup file contents into a `VocabLabelsResponse` object
- Added graceful exception handling for network errors encountered when requesting external vocabs on startup (cogatlas)

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [x] Tests have been added

For bug fixes:

### DIFF
--- a/app/api/crud.py
+++ b/app/api/crud.py
@@ -275,11 +275,11 @@ async def get_controlled_term_attributes() -> list:
     return results_list
 
 
-async def get_term_labels_for_cogatlas(
-    term_labels_path: Path,
+async def get_term_labels_for_vocab(
+    term_labels_path: Path, vocabulary_name: str, namespace_prefix: str
 ) -> VocabLabelsResponse:
     """
-    Returns the term-label mappings along with the vocabulary namespace details for the Cognitive Atlas Task vocabulary.
+    Returns the term-label mappings along with the vocabulary namespace details for the specified vocabulary.
 
     Returns
     -------
@@ -288,8 +288,8 @@ async def get_term_labels_for_cogatlas(
     term_labels = util.load_json(term_labels_path)
 
     return VocabLabelsResponse(
-        vocabulary_name="Cognitive Atlas Tasks",
-        namespace_url=util.CONTEXT["cogatlas"],
-        namespace_prefix="cogatlas",
+        vocabulary_name=vocabulary_name,
+        namespace_url=util.CONTEXT[namespace_prefix],
+        namespace_prefix=namespace_prefix,
         term_labels=term_labels,
     )
--- a/app/api/routers/attributes.py
+++ b/app/api/routers/attributes.py
@@ -13,9 +13,18 @@ async def get_term_labels_for_vocab(
 ):
     """When a GET request is sent, return a dict containing the name, namespace info, and all term ID-label mappings for the vocabulary of the specified variable."""
     if data_element_URI is DataElementURI.assessment:
-        return await crud.get_term_labels_for_cogatlas(
-            term_labels_path=request.app.state.cogatlas_term_lookup_path
+        return await crud.get_term_labels_for_vocab(
+            term_labels_path=request.app.state.cogatlas_term_lookup_path,
+            vocabulary_name="Cognitive Atlas Tasks",
+            namespace_prefix="cogatlas",
         )
+    else:
+        if data_element_URI is DataElementURI.diagnosis:
+            return await crud.get_term_labels_for_vocab(
+                term_labels_path=request.app.state.snomed_term_lookup_path,
+                vocabulary_name="SNOMED CT",
+                namespace_prefix="snomed",
+            )
 
 
 @router.get("/{data_element_URI}")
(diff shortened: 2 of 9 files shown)


rmanaem (approved):
Thanks for the PR @alyssadai! Everything looks good. I just left a suggestion and a question, please take a look and then :cook:

app/api/crud.py
```diff
@@ -275,11 +275,11 @@ async def get_controlled_term_attributes() -> list:
     return results_list
 
 
-async def get_term_labels_for_cogatlas(
-    term_labels_path: Path,
+async def get_term_labels_for_vocab(
+    term_labels_path: Path, vocabulary_name: str, namespace_prefix: str
```
surchs, line 279: :tada:

app/api/routers/attributes.py
```diff
@@ -12,10 +12,19 @@ async def get_term_labels_for_vocab(
...
-    if data_element_URI is DataElementURI.assessment:
-        return await crud.get_term_labels_for_cogatlas(
-            term_labels_path=request.app.state.cogatlas_term_lookup_path
+    if data_element_URI == DataElementURI.assessment:
+        return await crud.get_term_labels_for_vocab(
+            term_labels_path=request.app.state.cogatlas_term_lookup_path,
+            vocabulary_name="Cognitive Atlas Tasks",
+            namespace_prefix="cogatlas",
         )
+    else:
```
surchs, line 21: I think this `else` is not needed here.

It's a (relatively) new language feature, but you could also take a look at https://peps.python.org/pep-0634/#the-match-statement

app/api/routers/attributes.py
```diff
@@ -30,6 +39,9 @@ async def get_terms(
 
     if data_element_URI == DataElementURI.assessment:
         term_labels_path = request.app.state.cogatlas_term_lookup_path
+    else:
```
surchs, line 42: is there are reason you add the `else` here?

surchs (commented):
I somehow missed the previous review. Just adding my comments here in case it helps

app/api/routers/attributes.py
alyssadai: Hm, I think you're right that it's not actually needed!

app/api/utility.py
```diff
@@ -369,26 +369,38 @@ def fetch_and_save_cogatlas(output_path: Path):
...
+                """
+            )
+            # Use backup copy of the raw vocabulary JSON
+            vocab = load_json(BACKUP_VOCAB_DIR / "cogatlas_task.json")
+    except httpx.NetworkError as exc:
         warnings.warn(
-            f"""
-            The API was unable to fetch the Cognitive Atlas task vocabulary (https://www.cognitiveatlas.org/tasks/a/) from the source and will default to using a local backup copy of the vocabulary instead.
+            f""""
+            Fetching of the Cognitive Atlas task vocabulary (https://www.cognitiveatlas.org/tasks/a/) from the source failed due to a network error.
```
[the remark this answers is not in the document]
alyssadai: I don't think so! `NetworkError` falls under the `RequestError` umbrella, which is separate from `HTTPStatusError` (which is what's raised by the `raise_for_status` call suggested by your other comment; see https://www.python-httpx.org/quickstart/#exceptions / https://www.python-httpx.org/exceptions/#exception-classes). `NetworkError` specifically includes problems establishing a network connection, so this `except` allows the API to still start up even if the user isn't connected to the internet. `raise_for_status`/`HTTPStatusError` covers cases where the request has gone through, whereas to my understanding, when `NetworkError`s arise, there's no response attribute (and so no status code).

That said, I can try to rewrite this bit to except _both_ `HTTPStatusError` (using `raise_for_status()`) and `NetworkError`. I think it should give me the same behaviour as the code currently.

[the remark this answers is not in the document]
alyssadai: So after some testing, I think I will keep the current syntax for now as switching to using `raise_for_status()` also requires more complicated mocking of the `httpx` functions in tests, which I think is not necessarily worthwhile for now as checking the status code directly serves the same purpose.

Happy to revisit this in the future if we need more complex exception handling/flow control!

PR #229, opened 2023-11-24
[FIX] Make session optional in SPARQL query template

### LINKED ISSUE #134: API returns a list of multiple nulls for `diagnosis` when healthy controls are selected
Example for one subject-session:
```json
     {
        "sub_id": "sub-10189",
        "session_id": "ses-nb01",
        "num_sessions": "1",
        "age": "49.0",
        "sex": "http://purl.bioontology.org/ontology/SNOMEDCT/248153007",
        "diagnosis": [
          null,
          null,
          null,
          null,
          null,
          null,
          null,
          null,
          null,
          null,
          null,
          null,
          null,
          null,
          null
        ],
        "subject_group": "purl:NCIT_C94342",
        "assessment": [
          "https://www.cognitiveatlas.org/task/id/trm_4f2453b806fe1",
          "https://www.cognitiveatlas.org/task/id/trm_4c8991e6e8597",
          "https://www.cognitiveatlas.org/task/id/trm_4da88a2a63d97",
          "https://www.cognitiveatlas.org/task/id/trm_4d559bcd67c18",
          "https://www.cognitiveatlas.org/task/id/trm_4c898a680e424"
        ],
        "image_modal": [
          "http://purl.org/nidash/nidm#T1Weighted",
          "http://purl.org/nidash/nidm#FlowWeighted",
          "http://purl.org/nidash/nidm#DiffusionWeighted"
        ],
        "session_file_path": "/home/adai/neurobagel/projects/bulk_annotations/inputs/openneuro-jsonld/ds000030/sub-10189"
      },
```

This most likely is happening because we are using `set()` to get the list of unique values for an attribute in the response from the graph, which treats `NaN`s as non-identical (see https://stackoverflow.com/questions/26245862/reducing-pandas-series-with-multiple-nan-values-to-a-set-gives-multiple-nan-valu)
https://github.com/neurobagel/api/blob/e58bec4443890308f9903401ef150be8d7c001f3/app/api/crud.py#L157-L169

So, when there are many records per subject-session due to the combination of selected variables in the graph, all instances of `null` values are returned in the resulting set.

### LINKED ISSUE #225: Update query template to make session optional
Datasets with imaging data that cannot be modeled by the CLI end up with JSONLDs in which subjects have no (imaging) session info. Since the query template used in the API assumes that all subjects have at least one session (see https://github.com/neurobagel/api/blob/18d5d95ecf8ae6c2ee4b56cbe7a279ef684a8498/app/api/utility.py#L185-L188), these datasets are never matched by any query sent using the API/query tool (e.g., this happens for 7/340 datasets in the `open_neuro` graph).

TODO:
- [ ] Update the API query template to match subjects without any imaging sessions (i.e., those with only phenotypic data)
- [ ] As part of this issue or soon after, probably move our data model limitations page to the official docs
  - [ ] and describe what happens when you are querying for # sessions

- See also https://github.com/neurobagel/planning/issues/54

alyssadai:
- Closes #225
- Closes #134
- Closes https://github.com/neurobagel/planning/issues/79

- make session-related attributes optional for subjects in the SPARQL query template (so that it matches subjects who have phenotypic info but no [imaging] sessions**, since currently sessions aren't created for phenotypic-only data)
- since subjects with only phenotypic data don't have a `session_id` (but the `subject_data` field in the response should have a separate entry per instance of subject data), update the pandas grouper used when processing the graph response to recognize `null` session ID when grouping the data by subject + session
  - otherwise, the `"subject_data"` field for these datasets will be an empty list `[]`
- for subject session-level attribute keys in the response that (theoretically) can be a list, `diagnosis`, `assessment`, etc., use `Series.unique()` instead of `set()` on the column values in the response dataframe to get the unique values per subject-session. This change is needed because `set()` does not work on multiple `null` values - meaning they are _all_ returned, e.g., `"diagnosis": [null, null, null, null]` 
  - see https://stackoverflow.com/questions/26245862/reducing-pandas-series-with-multiple-nan-values-to-a-set-gives-multiple-nan-valu

**note that these are subjects with only phenotypic data _in the graph_ (meaning they either don't have imaging data or the imaging data can't be modelled by neurobagel, so the CLI doesn't create a session layer for them)

### For reviewer:
1. For a basic sanity check the query template change recovers the 7/340 previously missing datasets for the OpenNeuro graph (see https://github.com/neurobagel/planning/issues/54), checkout the current branch and spin up the API locally with `NB_GRAPH_DB=open_neuro/query`), and run the following Python code:

```python
import httpx
response = httpx.get(url="http://127.0.0.1:8000/query/", timeout=30)
num_openneuro_datasets_NEW = len(response.json())  # should return 340

response = httpx.get(url="https://api.neurobagel.org/query/", timeout=30)
num_openneuro_datasets_OLD = len(response.json())  # should return 333
```

2. To see what the subject-level API response looks like for datasets that don't have any sessions in the graph, re-spin up the API pointed to `NB_GRAPH_DB=sysadmin_test/query` (also on `fairmount`; this db contains the 7 previously missing, pheno-only datasets as well as 1 'normal' dataset that has imaging sessions), make sure `NB_RETURN_AGG=false`, and run an empty query using the interactive docs.

3. If you want to manually check that this also fixes https://github.com/neurobagel/planning/issues/79, you can point your local API (with this branch's changes) in aggregate mode to the QPN or PPMI graphs (https://github.com/neurobagel/documentation/wiki/API#neurobagel-hosted-apis) and confirm that with an empty query, the `"dataset_total_subjects"` key and `"num_matching_subjects"` values for each dataset are now the same (I have already done this)

## Checklist

- [ ] PR has an interpretable title with a prefix (`[ENH]`, `[FIX]`, `[REF]`, `[TST]`, `[CI]`, `[MNT]`, `[INF]`, `[MODEL]`, `[DOC]`) _(see https://neurobagel.org/contributing/pull_requests for more info)_
- [ ] PR links to GitHub issue with mention `Closes #XXXX`
- [ ] Tests pass
- [ ] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:

### DIFF
--- a/app/api/utility.py
+++ b/app/api/utility.py
@@ -183,15 +183,17 @@ def create_query(
                     nb:hasLabel ?dataset_name;
                     nb:hasSamples ?subject.
             ?subject a nb:Subject;
-                    nb:hasLabel ?sub_id;
-                    nb:hasSession ?session;
-                    nb:hasSession/nb:hasAcquisition/nb:hasContrastType ?image_modal.
-            ?session nb:hasLabel ?session_id.
+                    nb:hasLabel ?sub_id.
             OPTIONAL {{
-                ?dataset_uuid nb:hasPortalURI ?dataset_portal_uri.
+                ?subject nb:hasSession ?session;
+                         nb:hasSession/nb:hasAcquisition/nb:hasContrastType ?image_modal.
+                ?session nb:hasLabel ?session_id.
+                OPTIONAL {{
+                    ?session nb:hasFilePath ?session_file_path.
+                }}
             }}
             OPTIONAL {{
-                ?session nb:hasFilePath ?session_file_path.
+                ?dataset_uuid nb:hasPortalURI ?dataset_portal_uri.
             }}
             OPTIONAL {{
                 ?subject nb:hasAge ?age.
@@ -211,9 +213,11 @@ def create_query(
             {{
                 SELECT ?subject (count(distinct ?session) as ?num_sessions)
                 WHERE {{
-                    ?subject a nb:Subject;
-                            nb:hasSession ?session.
-                    ?session nb:hasAcquisition/nb:hasContrastType ?image_modal.
+                    ?subject a nb:Subject.
+                    OPTIONAL {{
+                        ?subject nb:hasSession ?session.
+                        ?session nb:hasAcquisition/nb:hasContrastType ?image_modal.
+                    }}
                     {session_level_filters}
                 }} GROUP BY ?subject
             }}
--- a/app/api/crud.py
+++ b/app/api/crud.py
@@ -153,18 +153,18 @@ async def get(
             else:
                 subject_data = (
                     group.drop(dataset_cols, axis=1)
-                    .groupby(by=["sub_id", "session_id"])
-                    .agg(
+                    # TODO: Switch back to dropna=True once phenotypic sessions are implemented, as all subjects will have at least one non-null session ID
+                    .groupby(by=["sub_id", "session_id"], dropna=False).agg(
                         {
                             "sub_id": "first",
                             "session_id": "first",
                             "num_sessions": "first",
                             "age": "first",
                             "sex": "first",
-                            "diagnosis": lambda x: list(set(x)),
+                            "diagnosis": lambda x: list(x.unique()),
                             "subject_group": "first",
-                            "assessment": lambda x: list(set(x)),
-                            "image_modal": lambda x: list(set(x)),
+                            "assessment": lambda x: list(x.unique()),
+                            "image_modal": lambda x: list(x.unique()),
                             "session_file_path": "first",
                         }
                     )


app/api/utility.py
```diff
@@ -183,15 +183,17 @@ def create_query(
...
-            ?session nb:hasLabel ?session_id.
+                    nb:hasLabel ?sub_id.
             OPTIONAL {{
-                ?dataset_uuid nb:hasPortalURI ?dataset_portal_uri.
+                ?subject nb:hasSession ?session;
+                         nb:hasSession/nb:hasAcquisition/nb:hasContrastType ?image_modal.
+                ?session nb:hasLabel ?session_id.
+                OPTIONAL {{
+                    ?session nb:hasFilePath ?session_file_path.
+                }}
```
surchs, line 193: does that need to be a nested OPTIONAL? i.e. doesn't every session have a filePath?

app/api/utility.py
```diff
@@ -211,9 +213,11 @@ def create_query(
...
                 SELECT ?subject (count(distinct ?session) as ?num_sessions)
                 WHERE {{
-                    ?subject a nb:Subject;
-                            nb:hasSession ?session.
-                    ?session nb:hasAcquisition/nb:hasContrastType ?image_modal.
+                    ?subject a nb:Subject.
+                    OPTIONAL {{
+                        ?subject nb:hasSession ?session.
+                        ?session nb:hasAcquisition/nb:hasContrastType ?image_modal.
+                    }}
```
surchs, line 220: nice, does that end up counting to zero if the OPTIONAL doesn't exist?

surchs (approved):
very nice, thanks for addressing this!

:cook:

surchs:
Ah, I just saw your review suggestions. I took the easy way of just reading it. lmk if you'd prefer an actual test from someone else

app/api/utility.py
alyssadai: > doesn't every session have a filePath?

I think this is the case for all our graphs right now, but I kept this in since I noticed in our data model the file path attribute is technically optional:
https://github.com/neurobagel/bagel-cli/blob/4da00b6db4cce30d40f101c0c4e17be25db3828f/bagel/models.py#L52-L56

Once we support purely phenotypic sessions, I think this `OPTIONAL` will also become relevant. I've added a note on this in the related issue:
- https://github.com/neurobagel/bagel-cli/issues/225

app/api/utility.py
alyssadai: indeed!

PR #231, opened 2023-11-28
[FIX] Copy `vocab/` directory into container filesystem in Dockerfile

### LINKED ISSUE #230: `vocab/` directory not made available to container filesystem
Starting a container using the latest API Docker image fails with the following error:
```bash
FileNotFoundError: [Errno 2] No such file or directory: '/usr/src/vocab/backup_external/snomedct_disorder.json'
```

This is because we have not yet updated our Dockerfile to copy the top-level `vocab/` directory housing our backup JSON dumps of our external vocabs into the container filesystem, so the app cannot find the files (we didn't encounter this issue when we first added the CogAtlas JSON because the request to the source API always succeeded, so the backup file was never needed).

https://github.com/neurobagel/api/blob/25dd6aeb7d5800f9cbcf5c05ccac74541e1a88a7/Dockerfile#L9

alyssadai:
- Closes #230

- Fixes a bug where the SNOMED terms JSON dump cannot be found by the app on startup, because the `vocab/` directory doesn't exist in the container.

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,6 +7,7 @@ COPY ./requirements.txt /usr/src/app/requirements.txt
 RUN pip install --no-cache-dir --upgrade -r /usr/src/app/requirements.txt
 
 COPY ./app /usr/src/app
+COPY ./vocab /usr/src/vocab
 
 # NB_API_PORT, representing the port on which the API will be exposed, 
 # is an environment variable that will always have a default value of 8000 when building the image


surchs (approved):
:cook:

PR #255, opened 2024-01-16
🔄 synced file(s) with neurobagel/workflows

surchs:
synced local file(s) with [neurobagel/workflows](https://github.com/neurobagel/workflows).

<details>
<summary>Changed files</summary>
<ul>
<li>synced local directory <code>.github/workflows/</code> with remote directory <code>template_workflows/project_automation/</code></li><li>synced local directory <code>.github/workflows/</code> with remote directory <code>template_workflows/auto_release/</code></li><li>created local  <code>.autorc</code>  from remote  <code>template_configs/.autorc</code></li><li>created local  <code>.github/pull_request_template.md</code>  from remote  <code>template_prs/auto_release_repo_pr.md</code></li>
</ul>
</details>

---

This PR was created automatically by the [repo-file-sync-action](https://github.com/BetaHuhn/repo-file-sync-action) workflow run [#7536509328](https://github.com/neurobagel/workflows/actions/runs/7536509328)

PR #256, opened 2024-01-17
[CI] Update pre-commit CI config

### LINKED ISSUE #254: pre-commit CI fails on hooks with 'local' dependencies
See https://results.pre-commit.ci/run/github/567874792/1704917018.maEjX3MAQWuRwegTX2LaDw for example.

One of the pre-commit hooks we have in this repo is `docker-compose-check` which checks for valid compose files by running `docker compose config`.

This works fine when pre-commit is run locally, but errors out in CI because docker compose can't be found.

One option would be to use the `skip` config option in the `pre-commit-config.yaml` to skip `docker-compose-check` in CI:
https://pre-commit.ci/#configuration-skip
- in theor,y it should be possible to install docker compose in a github runner, but pre-commit CI itself doesn't actually use a GH workflow

Other options:
- A no-longer-maintained pre-commit GH action: https://github.com/pre-commit/action
  - superseded by pre-commit.ci
  - cons: seems like there's little to no configurability? although from piecing together some of the git history (e.g., https://github.com/pre-commit/action/pull/164), it looks like it doesn't make automatic pushes to PRs at least, which I think is what we want
- pre-commit.ci lite action: https://pre-commit.ci/lite
  - cons: only provides auto-fixing of PRs, which we don't need (and I can't find documentation on turning this option off/whether that is possible)

Relevant open issues in pre-commit.ci:
- https://github.com/pre-commit-ci/issues/issues/11
- https://github.com/pre-commit-ci/issues/issues/13

We should also ensure `autofix_prs` is False in this repo https://pre-commit.ci/#configuration-autofix_prs.

alyssadai:
- Closes #254 

Note: different Python repos require different pre-commit hooks, so `autofix: False` has to be set manually in each `.pre-commit-config.yaml`

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,3 +1,7 @@
+ci:
+  autofix_prs: false
+  skip: [docker-compose-check]
+
 repos:
   - repo: https://github.com/psf/black
     rev: 23.9.1


surchs (approved):
Looks good :cook: 

unrelated: why does this PR not have the new PR template (with the label-checkbox)?

alyssadai:
@surchs, because the PR with the new template is not merged yet, pending this fix of the pre-commit check failures:

https://github.com/neurobagel/api/pull/255/files

Issue #263, opened 2024-01-22
Split subject-level data list in API response into imaging and phenotypic session arrays


alyssadai:

surchs:
alternative
```
[
  {
    "dataset_uuid": "http://neurobagel.org/vocab/e0c7d08c-edcc-4c2a-816a-306878ed7be2",
    "dataset_name": "BIDS synthetic",
    "dataset_portal_uri": "https://github.com/bids-standard/bids-examples",
    "dataset_total_subjects": 5,
    "records_protected": false,
    "num_matching_subjects": 5,
    "subject_data": [
      {
        "sub_id": "sub-01",
        "session_id": "ses-01",
        "num_phenotypic_sessions": 2,
        "num_imaging_sessions": 2,
        "session_type": "http://neurobagel.org/vocab/ImagingSession",
        "age": null,
        "sex": null,
        "diagnosis": [
          null
        ],
        "subject_group": null,
        "assessment": [
          null
        ],
        "image_modal": [
          "http://purl.org/nidash/nidm#T1Weighted",
          "http://purl.org/nidash/nidm#FlowWeighted"
        ],
        "session_file_path": "/data/neurobagel/bagel-cli/bids-examples/synthetic/sub-01/ses-01"
      },
      {
        "sub_id": "sub-01",
        "session_id": "ses-01",
        "num_phenotypic_sessions": 2,
        "num_imaging_sessions": 2,
        "session_type": "http://neurobagel.org/vocab/PhenotypicSession",
        "age": 34.1,
        "sex": "http://purl.bioontology.org/ontology/SNOMEDCT/248152002",
        "diagnosis": [
          null
        ],
        "subject_group": "http://ncicb.nci.nih.gov/xml/owl/EVS/Thesaurus.owl#C94342",
        "assessment": [
          "https://www.cognitiveatlas.org/task/id/trm_57964b8a66aed",
          "https://www.cognitiveatlas.org/task/id/tsk_4a57abb949ece"
        ],
        "image_modal": [
          null
        ],
        "session_file_path": null
      },
```

Issue #262, opened 2024-01-22
Decide whether `"session_type"` in API response should be a URI or human-readable


alyssadai:
One consideration is that if a user wants to manually join the imaging + session level data available for a session based on the `session_id`, they will need to know how to read the `session_type`. In this case, a human-readable label "phenotypic" or "imaging" might be clearer.

alyssadai:
We have decided to keep the URI for `session_type` for now, as this best matches what we are currently returning for the other attributes.

PR #264, opened 2024-01-22
Support queries of session-level phenotypic attributes

### LINKED ISSUE #249: update query template to work with new session-level phenotypic data model
- [x] create local db with https://github.com/neurobagel/neurobagel_examples/blob/main/data-upload/pheno-bids-output/example_synthetic_pheno-bids.jsonld for testing new functionality
- [x] update query model: split query parameter for min # sessions into
    - min # pheno sessions
    - min # imaging sessions
- note: we won't have an "any" param cause it gets too complicated (it would need to be exclusive from min_num_pheno_sessions and min_num_imaging_sessions)
- note 2: default for either parameter cannot be 1, or this gets undesirably constrained when user doesn't fill out either field
- [x] update query template
- What do we return for a subject when they match the criteria?
  - Since the session ID of their phenotypic + imaging sessions can be different, we should return both separately + info for both (b/c otherwise, if user only sets an imaging data criteria, the matching sessions have no phenotypic attributes attached which is not useful)
  - [x] Separate entries for each session
  - [x] Per session entry, a new "session_type" key
- [x] in subject-level response, split `num_sessions` into phenotypic + imaging ses count
- [x] create new data model for session entry in `subject_data` response field
- [x] update docstrings of functions, remove TODOs
- [x] update tests

Reference query examples:

1. 2+ imaging sessions; imaging modality = T1
2. 2+ pheno sessions; age > 60; assessment = MoCA
3. 2+ imaging sessions; age > 60; assessment = MoCA
4. 2+ imaging sessions; 2+ pheno sessions; age > 60; assessment = MoCA

### LINKED ISSUE #251: Extend default timeout of `post_query_to_graph()`
The current default value is 5 s for some reason, which is even less than the httpx default (10 s).

We should change this to 30 s.

alyssadai:
- Closes #249 
- Closes #251 

See also:
- #262 
- #263

- Default timeout for POST request to graph extended
- `min_num_sessions` query parameter split up for phenotypic and imaging sessions
- New data model created for a specific session result in the API response
  - Includes new key `"session_type"` to differentiate phenotypic from imaging data
- Removed unnecessary manual ordering of variables returned from graph
- SPARQL query template updated to query for session-level attributes and count matching phenotypic sessions based on session-level filters
- Existing tests updated

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [x] Tests have been added

For bug fixes:


app/api/utility.py
```diff
@@ -133,61 +136,74 @@ def create_query(
...
+    if min_num_phenotypic_sessions is not None:
+        subject_level_filters += (
+            "\n"
+            + f"FILTER (?num_phenotypic_sessions >= {min_num_phenotypic_sessions})."
+        )
+    if min_num_imaging_sessions is not None:
+        subject_level_filters += (
+            "\n"
+            + f"FILTER (?num_imaging_sessions >= {min_num_imaging_sessions})."
+        )
```
surchs, line 186: should we move this section above pheno_filters now so the flow is more in line with the hierarchy (e.g. subject > pheno session - imaging session)?

app/api/utility.py
```diff
@@ -196,29 +212,55 @@ def create_query(
...
+                        }}
+                        OPTIONAL {{
+                            ?phenotypic_session nb:isSubjectGroup ?subject_group.
+                        }}
+                        OPTIONAL {{
+                            ?phenotypic_session nb:hasAssessment ?assessment.
+                        }}
+                    }}
+                    {phenotypic_session_level_filters}
+                }} GROUP BY ?subject
```
surchs, line 253: I think this is cleaner to wrap my head around than sticking both pheno and imaging sub-searches in the same `SELECT` statement - but just for my understanding: is that the only reason you made them into two separate sub-queries? Or is there some other aspect that makes this necessary and that I am missing?

surchs (approved):
Thanks @alyssadai, that's a pretty big PR. From reading through it all (most) makes a lot of sense to me and I added questions / suggestions where not. Take a look, but I think this is all good to go :cook:

tests/test_query.py
```diff
@@ -305,7 +305,8 @@ async def mock_get(
         sex,
         diagnosis,
         is_control,
-        min_num_sessions,
+        min_num_imaging_sessions,
+        min_num_phenotypic_sessions,
         assessment,
         image_modal,
     ):
```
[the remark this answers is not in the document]
alyssadai: Good idea. Tracked in https://github.com/neurobagel/api/issues/266

app/api/utility.py
alyssadai: Yes, I thought that two sub-queries (and sets of query filters) was more intuitive and easier to read.

From what I can tell, you can achieve functionally the same behaviour with a single `SELECT` statement!

PR #269, opened 2024-01-26
[TST] Refactor tests

### LINKED ISSUE #126: Write a test for behaviour of `RETURN_AGG` option
Currently, several of the tests in test_query.py use an aggregate query results object (`test_data()`) for testing.

We don't currently have a test that checks the structure of query results returned from the API depending on the set value of `RETURN_AGG`.

### Steps to implement
We could do one or both of
- ~testing the `create_query` function to check that the SPARQL query is wrapped with an aggregator statement when `RETURN_AGG=true`~
- mocking a small response from the graph (monkeypatching the `httpx` post function) to see if the graph results are parsed appropriately on the API side depending on `RETURN_AGG`

### LINKED ISSUE #266: Refactor mock get method for testing to accept return object
We keep using a `mocked_get()` function pretty much everywhere. A variant of this has already been refactored into the `conftest.py` file, so it probably makes sense to just have a single mock method but have it accept the return type as an argument. Something like so:

```python
def my_mock_get(expectedReturn):
  async def inner_mock(some, other, args):
    return expectedReturn
  return inner_mock
```

It seems that you can do something like this with fixtures in pytest using indirect parameterizations: https://docs.pytest.org/en/latest/example/parametrize.html#apply-indirect-on-particular-arguments

_Originally posted by @surchs in https://github.com/neurobagel/api/pull/264#discussion_r1462460857_

alyssadai:
- Closes #126
- Closes #266 

- Basic test added for structure of cohort query response when `RETURN_AGG=true`
- Turned mocked get function used across query tests into a few different fixtures, with variations based on the type of value/object to be returned by the mocked function
  - Specific return objects for the mocked function specified in test definitions via [indirect parametrization](https://docs.pytest.org/en/latest/example/parametrize.html#indirect-parametrization)
  - `mock_successful_get` (which depends on another fixture for the return object) kept a separate function to avoid overly complicated parametrization when writing tests (i.e., you cannot easily pass one fixture as a parameter to another fixture)
 

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -53,7 +53,7 @@ def test_data():
 
 @pytest.fixture
 def mock_post_query_to_graph():
-    """Mock post_query_to_graph function that returns toy data containing a dataset with no modalities for testing."""
+    """Mock post_query_to_graph function that returns toy aggregate data containing a dataset with no modalities for testing."""
 
     def mockreturn(query, timeout=5.0):
         return {
@@ -99,6 +99,20 @@ def mock_post_query_to_graph():
                 ]
             },
         }
+
+    return mockreturn
+
+
+@pytest.fixture
+def mock_query_matching_dataset_sizes():
+    """
+    Mock query_matching_dataset_sizes function that returns the total number of subjects for a toy dataset 12345.
+    Can be used together with mock_post_query_to_graph to mock both the POST step of a cohort query and the corresponding query for dataset size,
+    in order to test how the response from the graph is processed by the API (crud.get).
+    """
+
+    def mockreturn(dataset_uuids):
+        return {"http://neurobagel.org/vocab/12345": 200}
 
     return mockreturn
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -111,17 +111,25 @@ def mock_query_matching_dataset_sizes():
     in order to test how the response from the graph is processed by the API (crud.get).
     """
 
-    def mockreturn(dataset_uuids):
+    def _mock_query_matching_dataset_sizes(dataset_uuids):
         return {"http://neurobagel.org/vocab/12345": 200}
 
-    return mockreturn
+    return _mock_query_matching_dataset_sizes
 
 
 @pytest.fixture
 def mock_get_with_exception(request):
-    """Mock get function that raises a specified exception."""
+    """
+    Mock get function that raises a specified exception.
 
-    async def mockreturn(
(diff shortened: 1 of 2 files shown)


tests/conftest.py
```diff
@@ -138,7 +152,27 @@ async def mockreturn(
         assessment,
         image_modal,
     ):
-        return None
+        return request.param
+
+    return mockreturn
```
surchs, line 157: neat :tada:

tests/test_query.py
```diff
@@ -114,6 +113,7 @@ def test_get_valid_age_single_bound(
     assert response.json() != []
 
 
+@pytest.mark.parametrize("mock_get", [None], indirect=True)
```
surchs, line 116: phew. That's cool, but I see what you mean, bit tricky to "just read". I guess what this tells me is that you tell `mock_get` to have a `None` return here?

surchs: I wonder if we haven't gone a little over-complex here. Pytest seems to have much simpler ways to pass data to fixtures from tests: https://docs.pytest.org/en/6.2.x/fixture.html#using-markers-to-pass-data-to-fixtures . I believe that would be a lot more readable. Wdyt?

surchs (approved):
Very cool PR @alyssadai, thanks for figuring all of this out.

Looking over the double parameterization here is impressive, but also a bit confusing:
1. when I read the parameterization I wonder: what is parameterized? There's only one parameter
2. when I read the test itself where I am using the mock I wonder: what does this mock actually do? 

I think we could probably live with that, but my sense is that pytest offers a simpler way to achieve the same outcome: https://docs.pytest.org/en/6.2.x/fixture.html#using-markers-to-pass-data-to-fixtures 

Maybe you could take a look at that section (and the "Factories as fixtures" just after it) and see if you agree. 

Otherwise: :cook:

tests/test_query.py
alyssadai: After looking into this, sticking with the original indirect parametrization approach seems to be more intuitive in terms of syntax and from a test-writing perspective, while still helping us avoid re-mocking get functions inside individual tests. So proceeding with the original changes!

surchs:
:rocket: PR was released in [`v0.2.0`](https://github.com/neurobagel/api/releases/tag/v0.2.0) :rocket:

PR #286, opened 2024-03-08
[ENH] Added root endpoint with welcome message and API docs link

### LINKED ISSUE #199: Change root path for API to display an informative landing message/page
Currently, navigating to the root of the API (`/`) displays a `{"detail": "not found"}` message by default.

We want to update our path config in FastAPI to instead show a page that displays a simple welcome message and directs users to visit the API docs for more information should be sufficient, e.g.:

> Welcome to the Neurobagel REST API!
>
> Please visit the documentation at /docs to view available API endpoints.

(If `/docs` part could be made a link to the `<base-url>/docs`, that would also be cool but is not a necessity.)

samadpls:
- Closes #199 

- Created root endpoint with welcome message and API docs link

## Checklist

- [X] PR has an interpretable title with a prefix (`[ENH]`, `[FIX]`, `[REF]`, `[TST]`, `[CI]`, `[MNT]`, `[INF]`, `[MODEL]`, `[DOC]`) _(see https://neurobagel.org/contributing/pull_requests for more info)_
- [X] PR links to GitHub issue with mention `Closes #XXXX`
- [x] Tests pass
- [ ] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -27,6 +27,7 @@ app.add_middleware(
     allow_headers=["*"],
 )
 
+
 @app.get("/", response_class=HTMLResponse)
 def root():
     """
@@ -40,7 +41,8 @@ def root():
         </body>
     </html>
     """
-    
+
+
 @app.get("/favicon.ico", include_in_schema=False)
 async def favicon():
     """
--- a/app/main.py
+++ b/app/main.py
@@ -9,7 +9,7 @@ import uvicorn
 from fastapi import FastAPI
 from fastapi.middleware.cors import CORSMiddleware
 from fastapi.openapi.docs import get_redoc_html, get_swagger_ui_html
-from fastapi.responses import ORJSONResponse, RedirectResponse, HTMLResponse
+from fastapi.responses import HTMLResponse, ORJSONResponse, RedirectResponse
 
 from .api import utility as util
 from .api.routers import attributes, query
--- a/app/main.py
+++ b/app/main.py
@@ -9,7 +9,7 @@ import uvicorn
 from fastapi import FastAPI
 from fastapi.middleware.cors import CORSMiddleware
 from fastapi.openapi.docs import get_redoc_html, get_swagger_ui_html
-from fastapi.responses import ORJSONResponse, RedirectResponse
+from fastapi.responses import ORJSONResponse, RedirectResponse, HTMLResponse
 
 from .api import utility as util
 from .api.routers import attributes, query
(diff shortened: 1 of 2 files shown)


samadpls:
output:
image

app/main.py
```diff
@@ -28,6 +28,21 @@
...
+@app.get("/", response_class=HTMLResponse)
+def root():
+    """
+    Display a welcome message and a link to the API documentation.
+    """
+    return """
+    <html>
+        <body>
+            <h1>Welcome to the Neurobagel REST API!</h1>
+            <p>Please visit the <a href="/docs">documentation</a> to view available API endpoints.</p>
```
surchs, line 40: > If /docs part could be made a link to the <base-url>/docs, that would also be cool but is not a necessity

@alyssadai was this comment from #199 in reference to running the API behind a proxy?

surchs (approved):
Thanks @samadpls for taking this on, this looks good to me. I'd wait for @alyssadai to take a look at my question, but after that I think we'd be good to merge :cook:

app/main.py
alyssadai: Nope, I was thinking of a hyperlink that takes you to the API docs, to make it easier for a user to know where to go next when they hit this welcome page (most useful when no NGINX redirects have been set up). So this looks fine to me!

alyssadai (changes requested):
Hey @samadpls, thanks again for addressing this! The welcome message looks great and the link seems to work as expected.

Do you want to take a stab at adding a basic test for the new root endpoint response so that our code coverage check passes? https://github.com/neurobagel/api/blob/main/tests/test_app_events.py might be a good place for it for now, and you can use something like https://github.com/neurobagel/api/blob/87a0373e08ab6b6e57fa5e9ccfc0f4c9fdae7dd1/tests/test_attributes.py#L32-L42 as a reference (although note that the above is a test for invalid parameters, whereas in this case you just need to check that the response from `/` contains some part of the welcome message you would expect, and maybe a successful status code). 

Let me know if you have any questions!

tests/test_attributes.py
```diff
@@ -7,6 +7,17 @@
 from app.api import utility as util
 
 
+def test_root(test_app):
```
alyssadai, line 10:
```suggestion
def test_root(test_app, set_test_credentials):
```

Note that we have to set some toy credentials in the test environment (which we do via [this pytest fixture](https://github.com/neurobagel/api/blob/87a0373e08ab6b6e57fa5e9ccfc0f4c9fdae7dd1/tests/conftest.py#L14-L18) in conftest.py)  to ensure we don't get an error when starting up the test app. 

This is what is currently causing the test workflow to fail in the PR checks: https://github.com/neurobagel/api/actions/runs/8363565151/job/22896810427?pr=286#step:6:107

alyssadai (commented):
Thanks @samadpls for the test!

The tests are currently failing on the PR (see https://github.com/neurobagel/api/actions/runs/8363565151/job/22896810427?pr=286#step:6:107), I've suggested an addition below that should fix the problem.

Once the fix is applied, this should be good to merge.

surchs:
:rocket: PR was released in [`v0.2.0`](https://github.com/neurobagel/api/releases/tag/v0.2.0) :rocket:

PR #326, opened 2024-07-24
[FIX] Exclude sessions missing a queried property from matches

### LINKED ISSUE #324: [BUG] Healthy Control results are incorrect

- [X] I have searched the existing issues

### Expected Behavior

When I search for healthy control, I only find healthy controls

### Current Behavior

When I search for healthy controls e.g. in QPN, I get nonsensical results (everyone is a healthy control). The input .jsonld files look correct, so I assume it's a search issue and not a CLI issue.

### Error message

_No response_

### Environment

_No response_

### How to reproduce

- grab a .jsonld with healthy controls and patients / not-controls
- upload it to a graph
- search for healthy controls and compare to the expected numbers

### Anything else?

I think this may have to be moved to the node-API repo if it is in fact a query issue

alyssadai:
- Closes #324

- Use the `BOUND` function in subquery for matching phenotypic sessions to exclude any results where a variable being filtered for does not have a value ("unbound" results - these are included by default, leading to inaccurate counts of matches)
  - e.g., because all phenotypic attributes are optional, when we query for healthy controls, the default matching behaviour _includes_ any sessions that do not have the `isSubjectGroup` edge - which would be any session that has diagnosis info, since these two properties are mutually exclusive
- Make having a phenotypic session non-optional in query (since subjects should not end up in the graph if they don't have any phenotypic info)
- Minor refactoring of query template for efficiency

This PR also includes a small fix to the context of an auth behaviour test that assumes authentication is enabled (this test was failing for me locally b/c I had auth turned off in development).

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:

### DIFF
--- a/app/api/utility.py
+++ b/app/api/utility.py
@@ -212,13 +212,13 @@ def create_query(
         ?diagnosis ?subject_group ?num_matching_phenotypic_sessions ?num_matching_imaging_sessions ?session_id ?session_type ?assessment ?image_modal ?session_file_path
         WHERE {{
             ?dataset_uuid a nb:Dataset;
-                    nb:hasLabel ?dataset_name;
-                    nb:hasSamples ?subject.
+                nb:hasLabel ?dataset_name;
+                nb:hasSamples ?subject.
             ?subject a nb:Subject;
-                    nb:hasLabel ?sub_id;
-                    nb:hasSession ?session.
+                nb:hasLabel ?sub_id;
+                nb:hasSession ?session.
             ?session a ?session_type;
-                    nb:hasLabel ?session_id.
+                nb:hasLabel ?session_id.
             OPTIONAL {{
                 ?session nb:hasAcquisition/nb:hasContrastType ?image_modal.
                 OPTIONAL {{?session nb:hasFilePath ?session_file_path.}}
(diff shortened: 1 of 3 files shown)


app/api/utility.py
```diff
@@ -217,24 +232,21 @@ def create_query(
...
+                    OPTIONAL {{?phenotypic_session nb:isSubjectGroup ?subject_group.}}
+                    OPTIONAL {{?phenotypic_session nb:hasAssessment ?assessment.}}
+
                     {phenotypic_session_level_filters}
                 }} GROUP BY ?subject
             }}
             {{
                 SELECT ?subject (count(distinct ?imaging_session) as ?num_matching_imaging_sessions)
                 WHERE {{
-                    ?subject a nb:Subject.
```
rmanaem, line 237: Are we removing this cause it's unnecessary to do since the only nodes that have a `nb:hasSession` edge to nodes of class `nb:ImagingSession` are of class `nb:Subject`?

app/api/utility.py
```diff
@@ -261,6 +273,7 @@ def create_query(
             + "} GROUP BY ?dataset_uuid ?dataset_name ?dataset_portal_uri ?sub_id ?image_modal"
         )
 
+    print(query_string)
```
rmanaem, line 276: Left over from debugging?

rmanaem (approved):
Thanks for the PR @alyssadai!
I left two comments, please take a look and then :cook:

app/api/utility.py
alyssadai: This statement can be removed because the main query (which comes first) already filters for subjects as being of type `nb:Subject`

https://github.com/neurobagel/api/blob/0ac0d4297d0942e555dadd066555540341cf6156/app/api/utility.py#L196-L202

so in subsequent subqueries we should be able to simply reuse the resulting `?subject` without having to repeat the type check again! This just makes the query a bit more efficient.

Issue #283, opened 2024-02-28
Write integration test that confirms the SPARQL query works against a current graph


surchs:
#282 is a good example of a PR that looks good but to be completely sure a reviewer would have to
- clone the PR branch
- spin up a local graph
- configure the graph (although that is easier now with https://github.com/neurobagel/recipes/pull/19)
- put data into the local graph
- either spin up the local PR branch directly and connect to the Graph by hand OR
- build a new docker image from the local PR branch, edit the docker-compose.yml and then spin up the local branch API with the graph

That's silly work and a bot should do it. As part of a test. We should be most of the way there already, but the rest is probably still a good bit of work. 

The crucial part of the issue entails testing that the query template returns expected results, rather than the content of the query string itself. Specifically, this may requiring construct test queries, and then asserting over responses from the graph.

This would help catch issues such as:
- https://github.com/neurobagel/api/issues/324
- https://github.com/neurobagel/api/issues/331

PR #349, opened 2024-10-10
[FIX] Ensure pipeline variables are returned from the graph in an aggregate query

### LINKED ISSUE #348: [BUG] Aggregate API responses always contain empty dict for `available_pipelines`

- [x] I have searched the existing issues

### Expected Behavior

When subject data in the graph contains pipeline metadata, an aggregate mode query response for the dataset should include a non-empty dictionary for `available_pipelines`.

### Current Behavior

`available_pipelines` is always an empty dictionary `{}`.

This is because the aggregation statement in the SPARQL query template has not yet been updated to return pipeline variables, and similarly, the API-side handling of the graph response for an aggregate query currently does not consider these variables.
https://github.com/neurobagel/api/blob/c150b37c659ffd16fe3d38d982f09ccd5edf996c/app/api/utility.py#L306-L315

At the same time, the value of `available_pipelines` is always initialized as an empty dict:
https://github.com/neurobagel/api/blob/c150b37c659ffd16fe3d38d982f09ccd5edf996c/app/api/crud.py#L171
https://github.com/neurobagel/api/blob/c150b37c659ffd16fe3d38d982f09ccd5edf996c/app/api/crud.py#L290

And so this is what is currently always returned in aggregate mode.

### Error message

_No response_

### Environment

- OS:
- Python/Node version:

### How to reproduce

Install and launch the API from source, in aggregate mode, and point it to a graph containing data from fhttps://github.com/neurobagel/neurobagel_examples/tree/main/data-upload/pheno-bids-derivatives-output.

### Anything else?

_No response_

alyssadai:
- Closes #348 

- added `pipeline_name` and `pipeline_version` to list of variables returned in the aggregate graph response
- updated API-side graph response handling to return pipeline info even in aggregate mode
- minor conditional rewriting for clearer logic

Note: In this case, some tests started failing after the graph response handling logic was updated, since the mocked aggregate response from the graph used in those tests no longer matched the expected format (good). However, we are not yet to write a test that would fail under the original bug conditions (bad), since this would require a graph with data to test against. See https://github.com/neurobagel/api/issues/283

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:

### DIFF
--- a/app/api/crud.py
+++ b/app/api/crud.py
@@ -168,7 +168,6 @@ async def get(
 
     response_obj = []
     dataset_cols = ["dataset_uuid", "dataset_name"]
-    dataset_available_pipeline_info = {}
     if not results_df.empty:
         for (dataset_uuid, dataset_name), group in results_df.groupby(
             by=dataset_cols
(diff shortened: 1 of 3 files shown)


surchs (approved):
Thanks @alyssadai for the bug find and fix. 

Agree that we should find a way to test e2e with a real graph so we can test unmocked query responses against our expectations. 

:cook:

PR #358, opened 2024-10-17
[REF] Split generic `/attributes` endpoints into attribute-specific routers

### LINKED ISSUE #354: Refactor generic `/attributes` router into attribute-specific routers
Given that our API is gradually expanding to accommodate requests for more specific metadata about specific attributes (e.g., versions of a specific pipeline), it makes sense to restructure the logic of existing `/attributes` routes to be more REST-compliant and intuitive for API consumers.

At the moment, we have two "top-level" subpaths in the `/attributes` router, which are:
- `/attributes/{data_element_uri}/`
- `/attributes/{data_element_uri}/vocab`

While the current level of abstraction makes it easy to accommodate new _attributes_ in the graph, this structure is less intuitive in terms of what each endpoint returns, and gets tricky to expand in terms of the _types_ of info returned. e.g., specific versions of the `nb:Pipeline` attribute:

- `/attributes/{data_element_uri}/versions` or even `/attributes/nb:Pipeline/versions` is not very elegant

Instead, it would be more helpful and maintainable to adopt a more conventional REST structure where top-level routers represent distinct entities (with potentially distinct operations and behaviours), and subpaths point to increasingly specific resources of those entities. This avoids path ambiguity for the end user cause by nesting everything under a single `/attributes` router, and makes the API more explicit and intuitive.

This can be seen in practice in the [Stardog](https://stardog-union.github.io/http-docs/#tag/Users) and [GraphDB](https://rdf4j.org/documentation/reference/rest-api/#tag/Repositories) API docs.

## Suggested path structure
- `/attributes/` -> to return all Neurobagel attributes in the data model (i.e., keeping the current behaviour)
- and then one top-level router for each attribute we support introspection of (most likely, all the ones with controlled terms as objects):
  - `/pipelines`
  - `/assessments`
  - `/diagnoses`
  - `/sex` (?)
- per attribute router, there should be:
  - a root path `/` that returns available instances in the graph
  - a vocab path `/vocab` that returns the possible labels and term URLs from the corresponding vocab

**NOTE** The f-API would also need to be updated to reflect this change

alyssadai:
- Closes #354 

- `GET /attributes/{data_element_uri}` and `GET /attributes/{data_element_uri}/vocab` endpoints removed
- Added two new routers:
  - `/assessments`
  - `/diagnoses`
- Created factory functions for the root `/` (i.e., get all instances in the graph) and `/vocab` subpaths of attribute-specific routers
- Refactored tests of removed endpoints into a new module `test_attribute_factory_routes.py`
- Vocab-specific variables for storing paths to external vocabulary term lookup files changed to a dictionary, for easier parametrization in factory routes

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [x] Tests have been added

For bug fixes:


app/api/routers/route_factory.py
```diff
@@ -0,0 +1,37 @@
+from fastapi import Request
+
+from .. import crud
+
+
+def create_get_instances_endpoint(
```
surchs, line 6: I think a better name is 
```suggestion
def create_get_instances_handler(
```
because the endpoint is already defined, you're supplying the handler function here.

app/api/routers/route_factory.py
```diff
@@ -0,0 +1,37 @@
...
+            request.app.state.vocab_lookup_paths[external_vocab]
+            if external_vocab is not None
+            else None
+        )
+        return await crud.get_terms(data_element_uri, term_labels_path)
+
+    return get_instances
+
+
+def create_get_vocab_endpoint(external_vocab: str, vocab_name: str):
```
surchs, line 25:
```suggestion
def create_get_vocab_handler(external_vocab: str, vocab_name: str):
```

app/api/routers/route_factory.py
```diff
@@ -0,0 +1,37 @@
...
+    """When a GET request is sent, return a dict containing the name, namespace info, and all term ID-label mappings for the vocabulary of the specified variable."""
+
```
surchs, line 27: The docstring should go into the wrapped function/handler here. And the outer function/decorator should get a different docstring.
```suggestion
    """When a GET request is sent, return a dict containing the name, namespace info, and all term ID-label mappings for the vocabulary of the specified variable."""

```
Something like "create the handler function"

surchs (approved):
Thanks @alyssadai, very cool PR. 

Couple of clarification comments, but otherwise :cook:

PR #364, opened 2024-10-24
[FIX] Allow only `"true"` or `None` for `is_control` query parameter

### LINKED ISSUE #332: Queries for `is_control=False` no longer work with restriction against unbound variables
This is really only a problem for the interactive docs, where setting `is_control` to `False` is an option.

See also this note:

https://github.com/neurobagel/api/blob/0bea618aec64dda22df43378236bbf2a30793704/app/api/utility.py#L191-L194

`hasSubjectGroup` can never be != `IS_CONTROL_TERM` because the only possible subject group at the moment _is_ healthy control. Otherwise, a subject would not have this property in the first place. So allowing users to query for a value of `False` will always return no results, violating the user expectations.

TODO: 
- [x] no longer allow `False` as a query value for `is_control`
- [x] update node API
- [ ] also update federation API (possibly additional issue for that)

alyssadai:
- Closes #332 

- `is_control` query param now accepts `str`, with allowed values being either `"true"` (case-insensitive) or `None` via custom validation
  - `str` used b/c using `Literal` to enforce allowed values of `True` or `None` doesn't play well in a query parameter (see comment in code)
  - `"true"` values transformed to boolean internally
- SPARQL query now only adds filter if `is_control=True` in query

Note: the default SPARQL query file hasn't changed in this case b/c it assumes no filters are applied

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [x] Tests have been added

For bug fixes:

### DIFF
--- a/app/api/models.py
+++ b/app/api/models.py
@@ -27,16 +27,20 @@ class QueryModel(BaseModel):
     # TODO: Check back if validating using a regex is too restrictive
     pipeline_version: constr(regex=VERSION_REGEX) = None
 
-    # TODO: If/once we migrate the n-API to Pydantic v2,
-    # instead of using the below approach to enforce a case-insensitive allowed value of "true" and convert it to a bool,
-    # we can probably use the new BeforeValidator method to do this directly in the query parameter definition.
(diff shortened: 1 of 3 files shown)


rmanaem (approved):
:cook:

PR #369, opened 2024-10-26
[FIX] Ensure non-agg API doesn't error out when all matches lack pipeline data

### LINKED ISSUE #368: Non-agg API response can't handle matching subjects lacking pipeline data
We need to ensure the presence of `NaNs` for `pipeline_name` and `pipeline_version` don't break things:

https://github.com/neurobagel/api/blob/99137363af22e2b81e212890d9cec291502cfd08/app/api/crud.py#L202-L220

alyssadai:
- Fixes #368

- Comments added to clarify steps in graph response reshaping
- Add mocked non-aggregated graph response where no subjects have pipeline info
  - Add test of API response for this query result case
- Handle `NaNs` for `pipeline_name` during graph response reshaping, to account for cases where there are no completed pipelines among matching sessions

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:

### DIFF
--- a/app/api/crud.py
+++ b/app/api/crud.py
@@ -238,6 +238,7 @@ async def get(
                     )
                     # NOTE: This expects a pd.Series and will not work on a pd.DataFrame
                     # (pd.DataFrame.reset_index() doesn't have a "name" arg)
+                    # See related https://github.com/pandas-dev/pandas/issues/55225
                     .reset_index(name="completed_pipelines")
                 )
 
@@ -247,12 +248,6 @@ async def get(
                     on=["sub_id", "session_id", "session_type"],
                     how="left",
                 )
-
-                # TODO: Remove - no longer needed?
-                # ensure that for sessions missing completed pipeline info, completed_pipelines is still a dict rather than null/nan
-                subject_data["completed_pipelines"] = subject_data[
-                    "completed_pipelines"
-                ].apply(lambda x: x if isinstance(x, dict) else {})
 
                 # TODO: Revisit this as there may be a more elegant solution.
                 # The following code replaces columns with all NaN values with values of None, to ensure they show up in the final JSON as `null`.
--- a/app/api/crud.py
+++ b/app/api/crud.py
@@ -208,8 +208,9 @@ async def get(
                             "session_type",
                             "pipeline_name",
                         ],
-                        # Keep NaNs to ensure that when there are no pipeline_name values in the query result,
-                        # we don't end up with an empty dataframe for pipeline_grouped_data
+                        # We cannot drop NaNs here because sessions without pipelines (i.e., with empty values for pipeline_name)
+                        # would otherwise be completely removed and in an extreme case where no matching sessions have pipeline info,
+                        # we'd end up with an empty dataframe.
                         dropna=False,
                     ).agg(
                         {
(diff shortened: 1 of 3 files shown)


app/api/crud.py
```diff
@@ -207,27 +208,37 @@ async def get(
...
+                            pname: pvers
+                            for pname, pvers in zip(
+                                x["pipeline_name"], x["pipeline_version"]
+                            )
+                            if not pd.isnull(pname)
+                        }
                     )
+                    # NOTE: This expects a pd.Series and will not work on a pd.DataFrame
+                    # (pd.DataFrame.reset_index() doesn't have a "name" arg)
+                    # See related https://github.com/pandas-dev/pandas/issues/55225
```
surchs, line 241: Can you explain how you could get a DataFrame here instead of a series? Maybe even linking
to the (soon to be closed) issue

app/api/crud.py
```diff
@@ -238,11 +249,6 @@ async def get(
                     how="left",
                 )
 
-                # ensure that for sessions missing completed pipeline info, completed_pipelines is still a dict rather than null/nan
-                subject_data["completed_pipelines"] = subject_data[
-                    "completed_pipelines"
-                ].apply(lambda x: x if isinstance(x, dict) else {})
```
surchs, line 244: why is this gone now?

tests/test_query.py
```diff
@@ -661,3 +669,67 @@ def test_integration_query_without_auth_succeeds(
...
+def test_missing_derivatives_info_handled_by_nonagg_api_response(
+    test_app,
+    mock_post_nonagg_query_to_graph,
+    mock_query_matching_dataset_sizes,
+    monkeypatch,
+    mock_auth_header,
+    set_mock_verify_token,
+):
+    """
+    Test that in the non-aggregated API mode, when all matching subjects lack pipeline information,
```
surchs, line 716: Maybe I'm missing it, but where is the test that

- has no pipeline info in the graph
- is running API in aggregated mode
- does not fail

? Because afaiu this is the test we would expect to catch our bug right now, yes?

surchs (approved):
Thanks @alyssadai for addressing this. I have a question about the test that protects against a regression here in the future. Otherwise this all looks nice 

:cook:

tests/test_query.py
alyssadai: That test doesn't exist. Do you think it's necessary to add it in this PR?  It wouldn't catch the current bug, so I'm not sure it would be super relevant here - rather, I think it should be addressed as part of https://github.com/neurobagel/api/issues/371.

The test that _would_ fail under the original bug conditions is the one you highlighted here, which is newly added, i.e.:
https://github.com/neurobagel/api/blob/b1733213de58911285b5f5fa7174910435e5e888/tests/test_query.py#L707-L735

surchs: aaah, thanks. I forgot that we had the bug when running in **non-agg** mode, not (as I falsely believed) in agg mode. So disregard my comment - agree with your answer

app/api/crud.py
alyssadai: Now that we're no longer dropping `NaN`s when we first group by `pipeline_name` here:
https://github.com/neurobagel/api/blob/b1733213de58911285b5f5fa7174910435e5e888/app/api/crud.py#L202-L213
all subject-sessions should now always be represented in the intermediate dataframes containing pipeline variables, regardless of whether they have pipeline info. So, any sessions missing pipeline info should already end up with a `{}` in the final `subject_data` (after this PR) without this extra processing. 

Previously, we needed this because any sessions without pipeline info would be dropped before reshaping the pipeline results, and so when we merged the dataframe with `completed_pipelines` info (`session_completed_pipeline_data`) back with the dataframe of all other session variables,
```
                subject_data = pd.merge(
                    subject_data.reset_index(drop=True),
                    session_completed_pipeline_data,
                    on=["sub_id", "session_id", "session_type"],
                    how="left",
                )
```
we would end up with some rows with empty values for `completed_pipelines`.

alyssadai: That said, it probably wouldn't hurt to keep this snippet in for now with a comment. I don't think it's necessary anymore, but might safeguard us against edge cases.

surchs: Ah, got it, thanks. I'd say we remove it (as you did)! Maybe the comment about not dropping NaNs can be amended instead to say: "we cannot drop NaNs here because sessions without pipelines would otherwise be completely removed and in an extreme case where none of the sessions have pipeline info, we'd end up with an empty dataframe"

Or something like that. Better to have a clear comment than to have dead code dangle

PR #397, opened 2024-12-20
[MNT] Replaced Cognitive Atlas with SNOMED

### LINKED ISSUE #390: Replace Cognitive Atlas vocab with SNOMED
The following places need to be updated:
- [x] vocab endpoint (including saved vocab file)
- [x] query template
- [x] `/assessments` endpoint
- [x] tests

rmanaem:
- Closes #390 

## Checklist

- [x] Tests pass
- [x] Checks pass
- [x] If the PR changes the SPARQL query template, the default Neurobagel query file has also been [regenerated](https://github.com/neurobagel/api?tab=readme-ov-file#the-default-neurobagel-sparql-query)

For new features:
- [ ] Tests have been added

For bug fixes:

## Summary by Sourcery

Switch from using the Cognitive Atlas to SNOMED CT for assessments.

### DIFF
--- a/app/api/routers/diagnoses.py
+++ b/app/api/routers/diagnoses.py
@@ -18,7 +18,7 @@ router.add_api_route(
     path="/vocab",
     endpoint=route_factory.create_get_vocab_handler(
         external_vocab=EXTERNAL_VOCAB,
-        vocab_name="SNOMED Disorder",
+        vocab_name="SNOMED CT Disorder",
         namespace_prefix="snomed",
     ),
     methods=["GET"],
--- a/app/api/routers/diagnoses.py
+++ b/app/api/routers/diagnoses.py
@@ -3,7 +3,7 @@ from fastapi import APIRouter
 from ..models import DataElementURI, VocabLabelsResponse
 from . import route_factory
 
-EXTERNAL_VOCAB = "snomed"
+EXTERNAL_VOCAB = "snomed_disorder"
 router = APIRouter(prefix="/diagnoses", tags=["diagnoses"])
 
 router.add_api_route(
@@ -17,7 +17,9 @@ router.add_api_route(
 router.add_api_route(
     path="/vocab",
     endpoint=route_factory.create_get_vocab_handler(
-        external_vocab=EXTERNAL_VOCAB, vocab_name="SNOMED CT"
+        external_vocab=EXTERNAL_VOCAB,
+        vocab_name="SNOMED Disorder",
+        namespace_prefix="snomed",
     ),
     methods=["GET"],
     response_model=VocabLabelsResponse,
--- a/app/api/utility.py
+++ b/app/api/utility.py
@@ -3,12 +3,9 @@
 import json
 import os
 import textwrap
-import warnings
 from collections import namedtuple
 from pathlib import Path
 from typing import Optional
-
-import httpx
 
 # Request constants
 EnvVar = namedtuple("EnvVar", ["name", "val"])
@@ -47,7 +44,6 @@ QUERY_HEADER = {
 }
 
 CONTEXT = {
-    "cogatlas": "https://www.cognitiveatlas.org/task/id/",
     "nb": "http://neurobagel.org/vocab/",
     "nbg": "http://neurobagel.org/graph/",  # TODO: Check if we still need this namespace.
     "ncit": "http://ncicb.nci.nih.gov/xml/owl/EVS/Thesaurus.owl#",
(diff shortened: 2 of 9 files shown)


app/api/routers/diagnoses.py
```diff
@@ -3,7 +3,7 @@
 from ..models import DataElementURI, VocabLabelsResponse
 from . import route_factory
 
-EXTERNAL_VOCAB = "snomed"
+EXTERNAL_VOCAB = "snomed_disorder"
```
surchs, line 6: why is the vocab file already renamed in the main branch? Is this from a daisy chained PR?

app/api/utility.py
```diff
@@ -451,54 +447,25 @@ def load_json(path: Path) -> dict:
...
-
-            Error: {exc}
-            """
-        )
-        # Use backup copy of the raw vocabulary JSON
-        vocab = load_json(BACKUP_VOCAB_DIR / "cogatlas_task.json")
+    vocab = load_json(BACKUP_VOCAB_DIR / "snomed_assessment.json")
 
-    term_labels = {term["id"]: term["name"] for term in vocab}
+    term_labels = {term["identifier"][7:]: term["label"] for term in vocab}
```
surchs, line 463: this double index confuses my brain. On a related note: it seems that we now have two snomed sub-vocabularies with different formats: 

**snomed disorder**:
```
{
    "sctid": "10007009",
    "preferred_name": "Coffin-Siris syndrome"
  },
```
and **snomed assessment**:

```
[
  {
    "identifier": "snomed:1231439001",
    "label": "Calgary Depression Scale for Schizophrenia"
  },
```
I think we should make both consistent and I think the assessment shape (identifier + label) is nicer

surchs (approved):
looks good to me @rmanaem, thanks for taking this on. I left some comments for you to look at. 

With those :cook:

app/api/routers/diagnoses.py
rmanaem: I believe it was called `snomedct_disorder` from the start.

app/api/utility.py
alyssadai: yeah, this discrepancy is related to the issue we discovered right before the break, where the vocab file formats currently used by the API and annotation tool are different. Specifically, the API (on `main`) uses vocab files that do not hardcode any vocab namespace prefixes

Here's the relevant issue https://github.com/neurobagel/planning/issues/216. I think the important decision would be whether we want to hardcode the prefixes or not into these files (see the "Considerations" section of the issue).

alyssadai: Also, agree with the double index being a bit hard to read! How about:
```suggestion
    term_labels = {term["identifier"].removeprefix("snomed:"): term["label"] for term in vocab}
```
should work for 3.9+

app/api/utility.py
```diff
@@ -47,7 +44,6 @@
 }
 
 CONTEXT = {
-    "cogatlas": "https://www.cognitiveatlas.org/task/id/",
     "nb": "http://neurobagel.org/vocab/",
     "nbg": "http://neurobagel.org/graph/",  # TODO: Check if we still need this namespace.
```
alyssadai, line 48: Since we're updating the context @neurobagel/dev should we just remove "nbg"? Don't think we use it anywhere, the CLI-generated contexts do not include this either.

tests/test_app_events.py
```diff
@@ -126,99 +125,8 @@ def test_stored_vocab_lookup_file_created_on_startup(
 ):
     """Test that on startup, a non-empty temporary lookup file is created for term ID-label mappings for the locally stored SNOMED CT vocabulary."""
     with test_app:
-        term_labels_path = test_app.app.state.vocab_lookup_paths["snomed"]
+        term_labels_path = test_app.app.state.vocab_lookup_paths[
+            "snomed_disorder"
```
alyssadai, line 129: 🍒: since we now create two SNOMED lookup files on startup instead of one, keyed on `snomed_disorder` and `snomed_assessment`, how about parametrizing this test to check that both paths exist?

tests/test_attribute_factory_routes.py
```diff
@@ -26,19 +26,19 @@ def test_get_instances_endpoint_with_vocab_lookup(
...
                         "type": "uri",
-                        "value": "https://www.cognitiveatlas.org/task/id/not_found_id",
+                        "value": "http://purl.bioontology.org/ontology/SNOMEDCT/not_found_id",
                     }
                 },
                 {
                     "termURL": {
                         "type": "uri",
-                        "value": "https://www.notanatlas.org/task/id/tsk_alz5hjlUXp4WY",
+                        "value": "http://notpurl.bioontology.org/ontology/SNOMEDCT/123456789",
```
alyssadai, line 41: for clarity, how about:
```suggestion
                        "value": "http://unknownvocab.org/123456789",
```

alyssadai (approved):
Thanks @rmanaem! Left a few 🍒 related to clarity. Otherwise, 🧑‍🍳

app/api/utility.py
surchs: agree

app/api/utility.py
rmanaem: Do we want to switch to identifier + label as part of this PR?

alyssadai: I would address that as part of https://github.com/neurobagel/planning/issues/216, since I think the format would benefit from some discussion

PR #408, opened 2025-02-05
[MNT] Remove trailing slashes from all paths

### LINKED ISSUE #399: Inconsistent trailing slashes in routes
In the [PR](https://github.com/neurobagel/api/pull/328) for:
- https://github.com/neurobagel/api/issues/327

we globally disabled redirection of trailing slashes, such that routes by default will now only work if they are accessed exactly as defined (e.g., if the `/query` route is defined without a trailing slash, then `/query/` will not work).

As part of the issue spec, we decided that we would globally define routes **without** the trailing slash for convenience and URLs, e.g.:

> api.neurobagel.org/query -> response
> api.neurobagel.org/query/ -> no response
> api.neurobagel.org/attributes/nb:Assessment -> response
> api.neurobagel.org/attributes/nb:Assessment/ -> no response

However, when we then refactored the generic `/attributes` route into variable-specific routes in:
- https://github.com/neurobagel/api/pull/358

we accidentally reintroduced the trailing slash into the variable routers. The result is that at the moment, we are not consistent with where we expect trailing slashes:

api.neurobagel.org/query -> works
api.neurobagel.org/query/ -> doesn't work
api.neurobagel.org/assessments -> doesn't work
api.neurobagel.org/assessments/ -> works

### Steps to implement
- [x] standardize the router definitions so that we do not expect `/` for the base of each route
- [x] test whether the root of app is reachable when path prefix has or lacks a trailing slash
- [x] check that custom favicon still works
- [ ] ensure that behaviour is predictable with generic stripped path prefix NGINX config (note default [merge_slashes](https://nginx.org/en/docs/http/ngx_http_core_module.html#merge_slashes) behaviour)

**NOTE:** The query tool and f-API may also be affected by this change

alyssadai:
- Closes #399 

- Remove trailing slashes from all defined paths except the root, which requires the trailing slash in the path definition (but is apparently indifferent to its inclusion)
- Add test of favicon path

Housekeeping:
- Rename `NB_NAPI_ROOT_PATH` -> `NB_NAPI_BASE_PATH` for consistency with query tool

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:

## Summary by Sourcery

Remove trailing slashes from API endpoints.

Bug Fixes:
- Prevent redirects for requests with trailing slashes on specific routes.

Tests:
- Update tests to reflect changes in route handling.

## Summary by Sourcery

Remove trailing slashes from API endpoints.

Bug Fixes:
- Prevent redirects for requests with trailing slashes on specific routes.

Tests:
- Add tests for favicon path.
- Update tests to reflect changes in route handling.

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -53,7 +53,7 @@ async def favicon():
     """
     Overrides the default favicon with a custom one.
 
-    NOTE: When the API is behind an NGINX reverse proxy that has a stripped path prefix (and root_path is defined),
+    NOTE: When the API is behind a reverse proxy that has a stripped path prefix (and root_path is defined),
     the custom favicon doesn't appear to work correctly for any API paths other than the docs,
     as the path in the favicon request isn't automatically adjusted to include the root path prefix.
     """
--- a/app/main.py
+++ b/app/main.py
@@ -52,6 +52,10 @@ def root(request: Request):
 async def favicon():
     """
     Overrides the default favicon with a custom one.
+
+    NOTE: When the API is behind an NGINX reverse proxy that has a stripped path prefix (and root_path is defined),
+    the custom favicon doesn't appear to work correctly for any API paths other than the docs,
+    as the path in the favicon request isn't automatically adjusted to include the root path prefix.
     """
     return RedirectResponse(url=favicon_url)
 
--- a/tests/test_routing.py
+++ b/tests/test_routing.py
@@ -34,7 +34,13 @@ def test_request_without_trailing_slash_not_redirected(
 
 @pytest.mark.parametrize(
     "invalid_route",
-    ["/query/", "/query/?min_age=20", "/attributes/nb:SomeClass/"],
+    [
+        "/query/",
+        "/query/?min_age=20",
+        "/attributes/",
+        "/assessments/",
+        "/assessments/vocab/",
+    ],
 )
 def test_request_with_trailing_slash_not_redirected(
     test_app, disable_auth, invalid_route
--- a/tests/test_routing.py
+++ b/tests/test_routing.py
@@ -1,7 +1,7 @@
 import pytest
 
 from app.api import crud
-from app.main import app
+from app.main import app, favicon_url
 
 
 @pytest.mark.parametrize(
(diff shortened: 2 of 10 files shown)


app/main.py
```diff
@@ -52,6 +52,10 @@ def root(request: Request):
 async def favicon():
     """
     Overrides the default favicon with a custom one.
+
+    NOTE: When the API is behind an NGINX reverse proxy that has a stripped path prefix (and root_path is defined),
```
surchs, line 56: should we just say "reverse proxy" now, instead of NGINX?

tests/test_routing.py
```diff
@@ -71,3 +77,23 @@ def test_docs_work_using_defined_root_path(
...
+
+@pytest.mark.parametrize(
+    "test_route,expected_status_code",
+    [("/favicon.ico", 307), ("/api/favicon.ico", 307)],
+)
+def test_favicon_reachable_with_defined_root_path(
+    test_app, monkeypatch, test_route, expected_status_code
+):
+    """
+    Test that when the API root_path is set to a non-empty string, the favicon is still reachable
```
surchs, line 90: aren't we calling this base_path now?

app/main.py
surchs: also: root_path -> base_path

surchs (approved):
Thanks @alyssadai, this looks good to me. You still have a couple references to `root_path` in there, so make sure to rename them to `base_path`. 

Otherwise I think this is good to go: :cook:

app/main.py
alyssadai: I use `root_path` to refer to the variable [defined in FastAPI](https://fastapi.tiangolo.com/advanced/behind-a-proxy/#providing-the-root_path) for a base path :)

tests/test_routing.py
alyssadai: as above - here `root_path` is referring to the name of the variable in FastAPI!

indeed, in our communications we refer to it as the base path

Issue #411, opened 2025-02-13
API does not error out/apply default for environment variables set to empty string


alyssadai:
For our required environment variables in the n-API, e.g., `NB_GRAPH_USERNAME`, we have a startup checks that they are set using syntax like the following:

https://github.com/neurobagel/api/blob/3718200213d81a07755bcd97365f6a5a2918dd97/app/main.py#L99-L100

However, this check only works if the variable **does not exist** in environment at all, and does not catch cases where the variable exists but is set to an empty string `""` (such as the [default docker compose behaviour](https://docs.docker.com/reference/compose-file/interpolation/) if an interpolated variable defined in `environment` is unset).

More sneakily, this docker compose behaviour also affects other environment variables that are _optional_ but which we expect to apply the n-API to apply a sensible default for, such as `NB_MIN_CELL_SIZE`. Since docker-compose.yml does not assign a default value, this variable is automatically set to `""`, meaning that the API no longer detects it as being unset. This results in unpredictable values for variables inside the n-API.

PR #416, opened 2025-02-21
[REF] Move environment variable handling to Pydantic `BaseSettings` class

### LINKED ISSUE #316: Env var in tests
- How do we test an `os.env` read? 
- How do we test that changing the env var actually results in the functions in the test having the new value?

For testing the time at which the env vars are read is at import time as opposed to at app start time/test runtime.

alyssadai:
- Closes #316 

- Replace all global variables for environment variables with a single `Settings` class using [pydantic-settings](https://docs.pydantic.dev/latest/concepts/pydantic_settings/)
- Use [pytest-env](https://github.com/pytest-dev/pytest-env) to set a subset of environment vars to non-default values in the environment before any tests run, to allow basic test that environment variables are read in correctly
  - otherwise, since environment variables are read in on app startup (meaning the TestClient used by tests already has a settings instance), modifying settings values are done through simply monkeypatching attributes
- Change dummy test of invalid credentials to a meaningful integration test

Housekeeping:
- Update some leftover Pydantic v1 syntax
- Update README badges

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:

## Summary by Sourcery

This pull request refactors the application to use Pydantic's `BaseSettings` for managing environment variables. This change improves code maintainability and type safety by centralizing environment variable configuration and validation within a dedicated settings class. The application now uses a settings object to access environment variables, and tests have been updated to reflect this change.

Enhancements:
- Migrated environment variable handling to Pydantic `BaseSettings` class for improved type safety and validation.
- Replaced direct access to environment variables with the new settings object.

Tests:
- Updated tests to use the new settings object instead of directly patching environment variables.
- Added a new test module to verify that settings are read correctly from environment variables.

## Summary by Sourcery

Refactors the application to use Pydantic's `BaseSettings` for managing environment variables, improving code maintainability and type safety. Tests have been updated to reflect this change.

Enhancements:
- Migrated environment variable handling to Pydantic `BaseSettings` class for improved type safety and validation.
- Replaced direct access to environment variables with a settings object.

Tests:
- Updated tests to use the new settings object instead of directly patching environment variables.
- Added a new test module to verify that settings are read correctly from environment variables.
- Added an integration test to verify that the app returns a 401 status code when given invalid credentials for the graph.

Chores:
- Updated some leftover Pydantic v1 syntax.

### DIFF
--- a/app/api/config.py
+++ b/app/api/config.py
@@ -1,5 +1,5 @@
 from pydantic import Field
-from pydantic_settings import BaseSettings, SettingsConfigDict
+from pydantic_settings import BaseSettings
 
 
 class Settings(BaseSettings):
@@ -7,13 +7,13 @@ class Settings(BaseSettings):
 
     # TODO: Make case-sensitive?
     # We don't want Pydantic errors to be raised when the environment variables are not set
-    model_config = SettingsConfigDict(validate_default=False)
+    # model_config = SettingsConfigDict(validate_default=False)
 
     root_path: str = Field(alias="NB_NAPI_BASE_PATH", default="")
     allowed_origins: str = Field(alias="NB_API_ALLOWED_ORIGINS", default="")
     # TODO: Figure out what to do about defaults for username/password
-    graph_username: str = Field(alias="NB_GRAPH_USERNAME", default=None)
-    graph_password: str = Field(alias="NB_GRAPH_PASSWORD", default=None)
+    graph_username: str | None = Field(alias="NB_GRAPH_USERNAME", default=None)
+    graph_password: str | None = Field(alias="NB_GRAPH_PASSWORD", default=None)
     graph_address: str = Field(alias="NB_GRAPH_ADDRESS", default="127.0.0.1")
     graph_db: str = Field(alias="NB_GRAPH_DB", default="repositories/my_db")
     graph_port: int = Field(alias="NB_GRAPH_PORT", default=7200)
@@ -22,9 +22,9 @@ class Settings(BaseSettings):
     # Double check how this is parsed from environment
     min_cell_size: int = Field(alias="NB_MIN_CELL_SIZE", default=0)
     auth_enabled: bool = Field(alias="NB_ENABLE_AUTH", default=True)
-    client_id: str = Field(alias="NB_QUERY_CLIENT_ID", default=None)
+    client_id: str | None = Field(alias="NB_QUERY_CLIENT_ID", default=None)
 
-    # TODO: Add query url and query header?
+    # TODO: Add query url as computed field and query header as constant
 
 
 settings = Settings()
--- a/app/api/config.py
+++ b/app/api/config.py
@@ -1,4 +1,4 @@
-from pydantic import Field
+from pydantic import Field, computed_field
 from pydantic_settings import BaseSettings
 
 
@@ -6,9 +6,6 @@ class Settings(BaseSettings):
     """Data model for settings."""
 
     # TODO: Make case-sensitive?
-    # We don't want Pydantic errors to be raised when the environment variables are not set
-    # model_config = SettingsConfigDict(validate_default=False)
-
     root_path: str = Field(alias="NB_NAPI_BASE_PATH", default="")
     allowed_origins: str = Field(alias="NB_API_ALLOWED_ORIGINS", default="")
     # TODO: Figure out what to do about defaults for username/password
@@ -24,7 +21,9 @@ class Settings(BaseSettings):
     auth_enabled: bool = Field(alias="NB_ENABLE_AUTH", default=True)
     client_id: str | None = Field(alias="NB_QUERY_CLIENT_ID", default=None)
 
-    # TODO: Add query url as computed field and query header as constant
+    @computed_field
+    def query_url(self) -> str:
+        return f"http://{self.graph_address}:{self.graph_port}/{self.graph_db}"
 
 
 settings = Settings()
--- a/app/api/config.py
+++ b/app/api/config.py
@@ -22,6 +22,7 @@ class Settings(BaseSettings):
     client_id: str | None = Field(alias="NB_QUERY_CLIENT_ID", default=None)
 
     @computed_field
+    @property
     def query_url(self) -> str:
         """Construct the URL of the graph store to be queried."""
         return f"http://{self.graph_address}:{self.graph_port}/{self.graph_db}"
--- a/app/api/config.py
+++ b/app/api/config.py
@@ -1,0 +1,17 @@
+from pydantic_settings import BaseSettings
+
+
+class Settings(BaseSettings):
+    """Data model for settings."""
+
+    nb_napi_base_path: str = ""
+    nb_api_allowed_origins: str = ""
+    nb_graph_username: str
+    nb_graph_password: str
+    nb_graph_address: str = "127.0.0.1"
+    nb_graph_db: str = "repositories/my_db"
+    nb_graph_port: int = 7200
+    # Double check how this is parsed from environment
+    nb_return_agg: bool = True
+    # Double check how this is parsed from environment
+    nb_min_cell_size: int = 0
--- a/app/api/config.py
+++ b/app/api/config.py
@@ -1,28 +1,29 @@
+"""Configuration environment variables for the API."""
+
 from pydantic import Field, computed_field
 from pydantic_settings import BaseSettings
 
 
 class Settings(BaseSettings):
-    """Data model for settings."""
+    """Data model for configurable API settings."""
 
-    # TODO: Make case-sensitive?
+    # NOTE: Environment variables are case-insensitive by default
+    # (see https://docs.pydantic.dev/latest/concepts/pydantic_settings/#case-sensitivity)
     root_path: str = Field(alias="NB_NAPI_BASE_PATH", default="")
     allowed_origins: str = Field(alias="NB_API_ALLOWED_ORIGINS", default="")
-    # TODO: Figure out what to do about defaults for username/password
     graph_username: str | None = Field(alias="NB_GRAPH_USERNAME", default=None)
     graph_password: str | None = Field(alias="NB_GRAPH_PASSWORD", default=None)
     graph_address: str = Field(alias="NB_GRAPH_ADDRESS", default="127.0.0.1")
     graph_db: str = Field(alias="NB_GRAPH_DB", default="repositories/my_db")
     graph_port: int = Field(alias="NB_GRAPH_PORT", default=7200)
-    # Double check how this is parsed from environment
     return_agg: bool = Field(alias="NB_RETURN_AGG", default=True)
-    # Double check how this is parsed from environment
     min_cell_size: int = Field(alias="NB_MIN_CELL_SIZE", default=0)
     auth_enabled: bool = Field(alias="NB_ENABLE_AUTH", default=True)
     client_id: str | None = Field(alias="NB_QUERY_CLIENT_ID", default=None)
 
     @computed_field
     def query_url(self) -> str:
+        """Construct the URL of the graph store to be queried."""
         return f"http://{self.graph_address}:{self.graph_port}/{self.graph_db}"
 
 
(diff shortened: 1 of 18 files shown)


app/api/config.py
```diff
@@ -0,0 +1,31 @@
...
+from pydantic import Field, computed_field
+from pydantic_settings import BaseSettings
+
+
+class Settings(BaseSettings):
+    """Data model for configurable API settings."""
+
+    # NOTE: Environment variables are case-insensitive by default
+    # (see https://docs.pydantic.dev/latest/concepts/pydantic_settings/#case-sensitivity)
+    root_path: str = Field(alias="NB_NAPI_BASE_PATH", default="")
```
surchs, line 12: Since it's `root_path` "inside" but `base_path` "outside", how about we make use of the `description` field of https://docs.pydantic.dev/latest/api/fields/#pydantic.fields.Field to add a nice reminder / description of what this guy does

app/api/config.py
```diff
@@ -0,0 +1,31 @@
...
+    min_cell_size: int = Field(alias="NB_MIN_CELL_SIZE", default=0)
+    auth_enabled: bool = Field(alias="NB_ENABLE_AUTH", default=True)
+    client_id: str | None = Field(alias="NB_QUERY_CLIENT_ID", default=None)
+
+    @computed_field
+    @property
+    def query_url(self) -> str:
+        """Construct the URL of the graph store to be queried."""
+        return f"http://{self.graph_address}:{self.graph_port}/{self.graph_db}"
+
```
surchs, line 29: :tada: neato

app/api/crud.py
```diff
@@ -10,9 +9,10 @@
 from fastapi import HTTPException, status
 
 from . import utility as util
+from .config import settings
 from .models import CohortQueryResponse, SessionResponse, VocabLabelsResponse
 
-ALL_SUBJECT_ATTRIBUTES = list(SessionResponse.__fields__.keys()) + [
+ALL_SUBJECT_ATTRIBUTES = list(SessionResponse.model_fields.keys()) + [
```
surchs, line 15: is this a pydantic v2 change?

app/api/crud.py
```diff
@@ -41,12 +41,12 @@ def post_query_to_graph(query: str, timeout: float = None) -> dict:
...
         response = httpx.post(
-            url=util.QUERY_URL,
+            url=settings.query_url,
             content=query,
             headers=util.QUERY_HEADER,
             auth=httpx.BasicAuth(
-                os.environ.get(util.GRAPH_USERNAME.name),
-                os.environ.get(util.GRAPH_PASSWORD.name),
+                settings.graph_username,
+                settings.graph_password,
```
surchs, line 49: this is really cool!

app/api/security.py
```diff
@@ -19,11 +16,11 @@
...
-    """Check if the CLIENT_ID environment variable is set."""
-    # The CLIENT_ID is needed to verify the audience claim of ID tokens.
-    if AUTH_ENABLED and CLIENT_ID is None:
+    """Check if the app client ID environment variable is set."""
+    # The client ID is needed to verify the audience claim of ID tokens.
+    if settings.auth_enabled and settings.client_id is None:
         raise RuntimeError(
-            "Authentication has been enabled (NB_ENABLE_AUTH) but the environment variable NB_QUERY_CLIENT_ID is not set. "
+            f"Authentication has been enabled ({Settings.model_fields['auth_enabled'].alias}) but the environment variable NB_QUERY_CLIENT_ID is not set. "
             "Please set NB_QUERY_CLIENT_ID to the client ID for your Neurobagel query tool deployment, to verify the audience claim of ID tokens."
```
surchs, line 24: not sure if this syntax is just used to print the variable name, but if so, https://docs.python.org/3/whatsnew/3.8.html#f-strings-support-for-self-documenting-expressions-and-debugging might be easier

app/main.py
```diff
@@ -1,8 +1,9 @@
 """Main app."""
 
-import os
 import warnings
 from contextlib import asynccontextmanager
+
+# from functools import lru_cache
```
surchs, line 6: Is this leftover? If not (e.g. for future use), please add a TODO flag and explain when we want to uncomment

app/main.py
```diff
@@ -13,9 +14,14 @@
...
 
 from .api import utility as util
+from .api.config import Settings, settings
 from .api.routers import assessments, attributes, diagnoses, pipelines, query
 from .api.security import check_client_id
 
+# @lru_cache
+# def get_settings():
+#     return Settings()
+
```
surchs, line 24: same here

tests/conftest.py
```diff
@@ -14,7 +13,7 @@ def test_app():
 @pytest.fixture
 def enable_auth(monkeypatch):
     """Enable the authentication requirement for the API."""
-    monkeypatch.setattr("app.api.security.AUTH_ENABLED", True)
+    monkeypatch.setattr(settings, "auth_enabled", True)
```
surchs, line 16: Is this still needed now? I thought with pytest-env we can now correctly control the env before each test is run

tests/test_app_events.py
```diff
@@ -1,66 +1,46 @@
...
 @pytest.mark.filterwarnings("ignore:.*NB_API_ALLOWED_ORIGINS")
 def test_start_app_without_environment_vars_fails(
-    test_app, monkeypatch, disable_auth
+    test_app, disable_auth, monkeypatch
 ):
     """Given non-existing username and password environment variables, raises an informative RuntimeError."""
-    monkeypatch.delenv(util.GRAPH_USERNAME.name, raising=False)
-    monkeypatch.delenv(util.GRAPH_PASSWORD.name, raising=False)
+    monkeypatch.setattr(settings, "graph_username", None)
+    monkeypatch.setattr(settings, "graph_password", None)
```
surchs, line 18: why do we have to monkeypatch this now? I expected that pytest-env allows us to just "mock" the env variable in from the perspective of the TestClient?

surchs: ah, I see. pytest-env only allows us to set one group of ENV variables for **all** test we're running. Hhm. How useful do you think that is then?

tests/test_app_events.py
```diff
@@ -1,66 +1,46 @@
...
-    """Given invalid environment variables for the graph, returns a 401 status code."""
-    monkeypatch.setenv(util.GRAPH_USERNAME.name, "something")
-    monkeypatch.setenv(util.GRAPH_PASSWORD.name, "cool")
-
-    def mock_httpx_post(**kwargs):
-        return httpx.Response(status_code=401)
-
-    monkeypatch.setattr(httpx, "post", mock_httpx_post)
-    response = test_app.get("/query", headers=mock_auth_header)
-    assert response.status_code == 401
```
surchs, line 42: I'm a little puzzled what this test was handling before? I think it would be good to have a test for what happens with bad env variables, but I guess that would have to be a real integration test (i.e. with a real un-mocked graph in the back)

tests/test_query.py
```diff
@@ -656,20 +647,33 @@ def test_query_without_token_succeeds_when_auth_disabled(
     assert response.status_code == 200
 
 
+@pytest.mark.integration
+@pytest.mark.filterwarnings("ignore:.*NB_API_ALLOWED_ORIGINS")
+def test_app_with_invalid_environment_vars(
```
surchs, line 652: ah OK, I see this test has moved here now.

tests/test_query.py
```diff
@@ -656,20 +647,33 @@ def test_query_without_token_succeeds_when_auth_disabled(
     assert response.status_code == 200
 
 
+@pytest.mark.integration
```
surchs, line 650: :tada:

tests/test_settings.py
```diff
@@ -0,0 +1,21 @@
+from app.api.config import Settings
+
+
+def test_settings_read_correctly():
+    """Ensure that settings are read correctly from environment variables, with correct types and default values."""
+    settings = Settings()
```
surchs, line 6: you mean that you're testing that the pytest.ini variables are correctly read here? Isn't that essentially testing pytest-env and pydantic-settings code? I.e. what would we have to mess up for this test to fail?

pytest.ini
```diff
@@ -3,3 +3,18 @@ markers =
     integration: mark integration tests that need the test graph to run
 ; Default to not running tests with the integration marker
 addopts = -m "not integration"
+; NOTE: The following environment variables are set to non-default values before any tests are run,
+; allowing us to actually test that user-provided values are parsed accurately from the environment
+; while avoiding issues related to import order in tests.
```
surchs, line 8:
```suggestion
; NOTE: Here we set the following environment variables to non-default values before any tests are run,
; so we can test that environment variables will be read accurately when they are set by the user.
; setting the env variables here avoids issues related to import order in tests.
```

surchs (approved):
Thanks @alyssadai. This looks a lot cleaner. I left some comments, mainly to make sure I understand correctly. Take a look.

Then :cook:

app/api/crud.py
alyssadai: yes!

app/api/security.py
alyssadai: @surchs could you provide an example of what you mean? I could be missing something but I don't really see how this syntax could be incorporated here...

The goal is to print the alias for the `auth_enabled` variable in the `Settings` class, which is `NB_ENABLE_AUTH`, without hard-coding it:
```
Authentication has been enabled (NB_ENABLE_AUTH) but the environment variable NB_QUERY_CLIENT_ID is not set. 
```

 Using the f-string syntax you provided, something like:

```python
f"Authentication has been enabled ({Settings.model_fields['auth_enabled'].alias=}) but the environment variable NB_QUERY_CLIENT_ID is not set. "
```
would produce
```
Authentication has been enabled (Settings.model_fields['auth_enabled'].alias='NB_ENABLE_AUTH') but the environment variable NB_QUERY_CLIENT_ID is not set.
```

which is not the desired behaviour.

(Side note: I just realized that `NB_QUERY_CLIENT_ID` should also no longer be hard-coded in this error message, which I will add to this PR)

tests/conftest.py
alyssadai: Yes, I'd prefer to keep this fixture for explicitness and modularity. `pytest-env` allows us to define variables in the env once before any tests are run (in case we want to test anything related to actual *reading* of env vars for example), but we cannot dynamically _modify_ the values of any env vars once the tests begin.

Having these fixtures that explicitly override the *parsed* env var values at the per-test level ensures that the settings are predicable for tests that rely on specific configurations (e.g., tests for specific behaviour when auth is enabled, disabled, etc.). Otherwise, they would always rely on a specific global starting set of env var values, resulting in less flexibility and transparency of the tests.

tests/test_app_events.py
alyssadai: > pytest-env only allows us to set one group of ENV variables for all test we're running

yes, at least from my understanding!

> How useful do you think that is then?

I think it's still useful because it means we can actually test how environment variables are read in by the app and stored in corresponding internal variables, which is different than testing the app's behaviour assuming a specific value for an internal variable (i.e., once an env var has already been read in). This is important because env vars are always read in as strings, but we have several variables we expect to be **non-strings** when referenced internally within the app (e.g., they should be stored as bools or ints). This is hard to test without pytest-env.

tests/test_settings.py
alyssadai: This test checks that variables set in the environment (handled pytest.ini & pytest-env) are stored in corresponding internal variables in the app as the type we desire, and with correct default values if env vars are unset.

It does test a little bit of Pydantic settings behaviour (mainly the default value part, maybe), but I think it moreso serves as a test that we have defined our Settings model according to app expectations/needs. e.g., the app internally expects `NB_ENABLE_AUTH` to be a boolean. If we failed to specify that this variable should be a `bool` in `Settings`, that would be caught by this test since env vars are always read in as strings by default.

tests/test_app_events.py
alyssadai: (just realized I drafted but never posted this comment)
this test also serves as a good precursor for:

- https://github.com/neurobagel/api/issues/411

in that we've seen there are cases where whether the default value has been applied is not obvious (if a variable has been unset vs. set to ""). We could confirm the actual behaviour in a test such as this one!

PR #529, opened 2026-02-05
[MNT] Setup uv

### LINKED ISSUE #527: Adopt uv for dependency management

surchs:
Update workflows

- Closes #527 

- use `uv` to handle dependency lockfile for the project
- move dependencies to pyproject.toml
- add dependabot support to update the uv.lock file (will have to be replaced by an updated dependabot sync template)

> [!NOTE]  
> We want to also have our local dev environments use the lockfile dependencies (to make sure local dev is the same as prod). Achieving this without installing `uv` locally is going to be quite hard / impossible with this setup. So our local setup guide will now include having uv available to handle local venv and installation.

> [!NOTE]  
> In the CI commands and the docker build, we do not use uv now. This is mainly to keep the diff limited. We can do this by first asking uv to build a `requirements.txt` lock file from the `uv.lock` file - and then using pip to install from the `requirements.txt` file as before.

## Checklist

- [x] Tests pass
- [x] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:

## Summary by Sourcery

Adopt uv-based Python project and dependency management and update CI to use it for installing and running development dependencies.

Enhancements:
- Define project metadata, runtime dependencies, and development dependency groups in a new pyproject.toml.
- Configure code style and type-checking tools (Black, isort, mypy) via pyproject.toml.

CI:
- Update test workflow to install uv and use it to sync dependencies instead of pip/requirements.txt.

Chores:
- Add uv.lock to track the resolved dependency set for reproducible environments.


README.md
```diff
@@ -115,25 +116,34 @@ The response should be a list of dictionaries containing info about datasets wit
...
-$ pip install -r requirements.txt
+uv sync --all-groups
 ```
 
+This will create a virtual environment (if one doesn't exist), install the package in editable mode, and install all dependency groups including dev dependencies.
+
 #### Launch the API
 
-To launch the API make sure you're in repository's main directory and in your environment where the dependencies are installed and environment variables are set.
+To launch the API make sure you're in the repository's main directory and that your `.env` file is configured.
```
alyssadai:
```suggestion
To launch the API, make sure you're in the repository root and that your [`.env` file has been configured](#set-the-environment-variables).
```

pyproject.toml
```diff
@@ -0,0 +1,57 @@
+[project]
+name = "neurobagel-api"
+dynamic = ["version"]
+requires-python = ">=3.10,<3.14"
+dependencies = [
+    "fastapi",
+    "pydantic>=2.10,<3",
+    "pydantic-settings",
+    "uvicorn",
+    "httpx",
```
alyssadai: these are core dependencies of `fastapi[standard]`
```suggestion
```

pyproject.toml
```diff
@@ -0,0 +1,57 @@
+[project]
+name = "neurobagel-api"
+dynamic = ["version"]
+requires-python = ">=3.10,<3.14"
+dependencies = [
+    "fastapi",
```
alyssadai:
```suggestion
    "fastapi[standard]",
```

.github/dependabot.yml
```diff
@@ -57,13 +57,8 @@ updates:
...
       - "dependencies"
 
-  # In repo settings we have configured dependabot to open PRs for security updates.
-  # Here we configure custom labels to be applied to security update PRs,
-  # while still preventing regular version update PRs.
-  # See https://docs.github.com/en/code-security/dependabot/dependabot-security-updates/configuring-dependabot-security-updates#overriding-the-default-behavior-with-a-configuration-file
-  # Note: this configuration only has an effect in repositories that have
-  # a requirements.txt file / use python / pip.
-  - package-ecosystem: "pip"
+# This is a temporary fix until we update the global sync workflow
```
alyssadai: good catch 👍
mind creating an issue for this in https://github.com/neurobagel/workflows and then adding it to the uv milestone so we don't forget?

alyssadai (changes requested):
Thanks a lot @surchs for implementing this and for your patience on my review! I've tested out `uv` locally using your branch and I think it will simplify dependency management considerably - happy to be rid of these restrictive requirements.txt files 🎉

Since we haven't updated dependencies in a while, I think the main issue atm is incompatibilities between locked package versions in `uv.lock` and the newer Python versions we want to support. Specifically:
- uv.lock currently specifies some pretty old versions of tools
- that don't have wheels for e.g., 3.12 or 3.13
- meanwhile, the new pyproject.toml allows 3.11-3.13
- but the test.yaml only runs tests against 3.10 (so the issues with newer Pythons are not caught in CI)

Left more details below and and a suggestion of how we can resolve this!

Two other high-level comments (with details below):
- I think we can include a bit more guidance in the README on using `uv`. Given how much `uv` abstracts away and the many commands available, I find that things like knowing which env I'm installing things into, how to add or update a dependency, etc. aren't obvious coming from pip + venv where each step is explicit. I think a few brief explainers of commands could go a long way in this regard (also to save ourselves confusion during this transition period). As mentioned below, we might want to eventually move this info into to our CONTRIBUTING.md
- I don't follow the NOTE in your PR description about not using `uv` in the test.yaml and Dockerfile. From what I can see those files have been updated anyways in this PR, but I find the temporary solution harder to review than if we were to fully use `uv` for the Docker image and test wf. Can we consider making those changes here to avoid a second PR that reverts the changes?

README.md
```diff
@@ -229,6 +243,30 @@ with open("docs/default_neurobagel_query.rq", "w") as file:
     file.write(create_query(return_agg=False))
 ```
 
+## We manage dependencies with `uv`
```
alyssadai:
```suggestion
#### Updating dependencies
```
I'd then move this entire section under `### Python`, at the end of the section

app/api/crud.py
```diff
@@ -287,7 +290,7 @@ async def query_records(
             }
 
             if settings.return_agg:
-                subject_data = "protected"
+                subject_data: str | list = "protected"
```
alyssadai: FYI, I think some of these type hints should no longer be required with the pydantic mypy plugin suggested in .pre-commit-config.yaml, but we can leave them in as well.

app/api/env_settings.py
```diff
@@ -13,9 +14,11 @@
...
 # This avoids having to thread the request object through every function that needs access to e.g., the context
 # and also makes it easier to mock the configuration during testing.
-CONTEXT = {}
-ALL_VOCABS = {}
-DATASETS_METADATA = {}
+# TODO: do something a bit more precise with these type hints.
+# For now they are here so mypy is happy.
+CONTEXT: dict[str, Any] = {}
+ALL_VOCABS: dict[str, Any] = {}
+DATASETS_METADATA: dict[str, Any] = {}
```
alyssadai: if we want to be more specific, I believe this is correct,
```suggestion
CONTEXT: dict[str, str] = {}
ALL_VOCABS: dict[str, list] = {}
DATASETS_METADATA: dict[str, dict] = {}
```

alyssadai (approved):
Thanks @surchs for the changes! 🎉 

Everything is looking good, and I can confirm the `uv` commands now work for me locally without dependency conflicts.

I left some final minor fixes and README suggestions, otherwise 🧑‍🍳

Before you merge, can you double check the failing [codecov check](https://github.com/neurobagel/api/pull/529/checks?check_run_id=68656858800) to see if there's anything we can do to make it green?

Issue #534, opened 2026-02-26
Add caching for vocabulary fetching to improve API startup reliability


SarthakJagota:
I noticed that vocabularies are fetched from GitHub during every API startup via `fetch_vocabularies()`.

This can slow down local development and introduces a dependency on external network availability. If GitHub is temporarily unavailable, the API may fail to start even when previously fetched vocabularies exist.

Proposed improvement:
- Introduce a lightweight caching layer to store fetched vocabularies locally
- Load vocabularies from cache on subsequent startups
- Add fallback to cache if fetching fails

This would improve startup performance, developer experience, and robustness without changing existing API behavior.

Happy to work on this if it aligns with the project direction.

PR #535, opened 2026-02-26
Add caching layer for vocabulary fetching during API startup

SarthakJagota:
## Summary
This PR introduces a lightweight caching mechanism for vocabulary fetching during API startup.

Currently, vocabularies are fetched from GitHub on every startup, which can slow initialization and create a dependency on network availability.

## Changes
- Added `cache.py` helper to store and load vocabularies locally
- Load vocabularies from cache when available
- Fetch and persist vocabularies when cache is missing
- Added fallback to cached vocabularies if fetching fails
- Added logging to clarify when cache vs remote fetch is used

## Impact
- Faster startup for local development
- Improved reliability when external services are unavailable
- No change to existing API behavior

## Notes
Cached vocabularies can be refreshed by removing the `.cache` directory.

## Summary by Sourcery

Introduce a local caching mechanism for vocabulary data during API startup to reduce reliance on remote fetching and improve startup performance.

New Features:
- Add a cache helper module to persist and load vocabularies from a local .cache directory.
- Load vocabularies from the local cache on startup when available instead of always fetching from GitHub.

Enhancements:
- Add a fallback to use cached vocabularies when remote fetching fails to improve startup robustness.
- Add logging to indicate whether vocabularies are loaded from cache or fetched remotely.

#534

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -327,3 +327,5 @@ docs/_book
 
 # TODO: where does this rule come from?
 test/
+
+.cache/
--- a/app/api/cache.py
+++ b/app/api/cache.py
@@ -1,0 +1,19 @@
+import json
+from pathlib import Path
+from typing import Any
+
+CACHE_DIR = Path(".cache")
+VOCAB_CACHE_FILE = CACHE_DIR / "vocabs.json"
+
+
+def save_vocab_cache(data: dict[str, Any]) -> None:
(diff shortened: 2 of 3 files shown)


surchs:
Thanks @SarthakJagota for opening #534 - we will discuss possible solutions on the issue and then outline an implementation. I imagine we'd want to have some basic cache invalidation and tests as well - so I'll close this PR for now

PR #539, opened 2026-03-04
Switch to Python 3.9+ type annotation syntax everywhere

NobleCoder69:
Replaced legacy typing syntax (Optional, Union, List, etc.)
with Python 3.9+ built-in generics and PEP 604 union types.

Examples:
- Optional[str] → str | None
- Union[A, B] → A | B
- List[str] → list[str]

All tests passing (145/145).

## Summary by Sourcery

Modernize type annotations to use Python 3.9+ syntax across API models and utility helpers.

Enhancements:
- Update Pydantic API models to use PEP 604 union types and built-in collection generics instead of legacy typing.Optional/Union forms.
- Align utility function signatures with the newer type annotation style for optional and collection parameters.

### DIFF
--- a/app/api/models.py
+++ b/app/api/models.py
@@ -1,7 +1,7 @@
 """Data models."""
 
 from enum import Enum
-from typing import Optional, Union
+# remove this line completely
 
 from fastapi.exceptions import HTTPException
 from pydantic import (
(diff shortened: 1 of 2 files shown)


alyssadai (approved):
Thanks for the contribution @NobleCoder69, and sorry it's taken so long for us to get around to reviewing it!

I tightened a few of the type annotations and removed a leftover comment. This is good to go! 🧑‍🍳

